PR #35, opened 2026-04-02
refactor(cli): limit local server autostart to local mode for TASK-116

xarmian:
## Summary
- require an explicitly configured `local` mode before the CLI auto-manages a Pad server
- skip implicit localhost behavior in CLI helper paths when the client is not configured
- add guardrail tests around local vs external server management

## Test plan
- `go build ./...`
- `GOCACHE=/tmp/docapp-go-cache go test ./...`
- `cd web && npm run build`
- `make install`


xarmian:
@codex

PR #62, opened 2026-04-04
fix: support searching by item ref (BUG-20)

xarmian:
## Summary
- Adds direct ref lookup (e.g. `TASK-5`, `BUG-8`) to the search endpoint so items are findable by their issue ID
- Sanitizes FTS5 queries by quoting tokens, preventing hyphens from being interpreted as NOT operators
- Deduplicates results when a ref match also appears in FTS results

## Test plan
- [ ] Search for an existing item by ref (e.g. `BUG-20`) → should appear as first result
- [ ] Search for a non-existent ref (e.g. `IDEA-999`) → should return no results (no error)
- [ ] Search for a regular text query (e.g. `OAuth`) → FTS still works as before
- [ ] Search for a ref where the item also has FTS matches → no duplicates
- [ ] Test in both CLI (`pad item search`) and web UI search

### DIFF
--- a/internal/store/search.go
+++ b/internal/store/search.go
@@ -45,31 +45,36 @@ func (s *Store) Search(params SearchParams) ([]SearchResult, error) {
 			refArgs = append(refArgs, params.Workspace)
 		}
 
-		row := s.db.QueryRow(refQuery, refArgs...)
-		var r SearchResult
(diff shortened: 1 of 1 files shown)


xarmian:
@codex

PR #92, opened 2026-04-13
fix: quick-win bug fixes (BUG-517, BUG-538, BUG-401)

xarmian:
## Summary

Three small, self-contained bug fixes:

- **BUG-517** — Apostrophes in workspace names caused extra hyphens in slugs (`dave-s-workspace` → `daves-workspace`). Fixed `slugify()` to strip apostrophes silently instead of converting them to hyphens. Added test cases.
- **BUG-538** — Moving an item to a different collection pushed a new entry to browser history. Now uses `replaceState: true` on the `goto()` call, matching the pattern used elsewhere in the codebase.
- **BUG-401** — `deriveItemClosure` incorrectly auto-closed items when all split children were done. Splitting work out doesn't mean the original is complete — changed `IsClosed` from `true` to `false` for the `split_into` closure kind.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/store/ ./internal/server/` passes (including new slugify test cases)
- [x] `npm run build` compiles cleanly
- [ ] Manually verify: create a workspace with an apostrophe in the name → slug has no extra hyphen
- [ ] Manually verify: move an item between collections → browser back button doesn't return to the old collection URL
- [ ] Manually verify: split an item, complete the split children → original item stays open

### DIFF
--- a/internal/server/item_lineage.go
+++ b/internal/server/item_lineage.go
@@ -176,9 +176,11 @@ func (s *Server) deriveItemClosure(item *models.Item, vis []string) (*models.Ite
 			RelatedItems: implementedBy,
 		}, nil
 	}
+	// NOTE: split_into does NOT auto-close the original item. Splitting work
+	// out doesn't mean the original is done — it still stands on its own.
 	if len(splitChildren) > 0 && allSplitChildrenDone {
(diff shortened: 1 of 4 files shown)


xarmian:
@codex

PR #99, opened 2026-04-13
feat: restore pad init as smart multi-step entry point

xarmian:
## Summary

- Adds a top-level `pad init` command that detects the current state and walks through each setup step as needed: configure → start server → bootstrap admin → authenticate → create/link workspace → install/update AI skill files
- Safe to re-run anytime — skips completed steps and shows a clean status summary when everything is healthy
- Refactors `pad workspace init` to use shared helpers (`ensureWorkspace`, `ensureSkills`) and adds a hint pointing users to `pad init` when prerequisites are missing

Ref: IDEA-499, PLAN-546

## Test plan

- [ ] Run `pad init` on an already-configured project — should show status summary
- [ ] Run `pad init` twice in a row — second run should be identical (idempotent)
- [ ] Run `pad workspace init` — should still work as before
- [ ] Verify `pad --help` shows `init` at the top level
- [ ] Verify `pad init --help` shows full usage
- [ ] Verify skill files are installed/updated during init flow

### DIFF
--- a/cmd/pad/init.go
+++ b/cmd/pad/init.go
@@ -13,6 +13,7 @@ import (
 
 	pad "github.com/xarmian/pad"
 	"github.com/xarmian/pad/internal/cli"
+	"github.com/xarmian/pad/internal/collections"
 	"github.com/xarmian/pad/internal/config"
 	"github.com/xarmian/pad/internal/models"
 )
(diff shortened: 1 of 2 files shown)


xarmian:
@codex

PR #103, opened 2026-04-14
feat: add admin password reset for other users

xarmian:
## Summary
- New `POST /api/v1/admin/users/{id}/reset-password` endpoint
- When email is configured: generates a reset token and sends a password reset email
- When email is not configured: generates a temporary password, sets it on the user, and invalidates all their sessions
- New `password_reset_by_admin` audit action with metadata (target user, method used)
- Frontend: "Reset Password" button in admin user edit panel with confirmation dialog and result display (shows email confirmation or temporary password in a copyable code block)

Closes TASK-555 — part of PLAN-552 (Admin Console Enhancement)

## Test plan
- [ ] Reset password for a user (without email configured) — verify temp password is returned and sessions invalidated
- [ ] Reset password for a user (with email configured) — verify reset email is sent
- [ ] Verify audit log entry is created with correct action and metadata
- [ ] Verify `go test ./...` passes
- [ ] Verify `npm run build` compiles cleanly

### DIFF
--- a/internal/models/activity.go
+++ b/internal/models/activity.go
@@ -28,7 +28,8 @@ const (
 	ActionOAuthLogin      = "oauth_login"
 	ActionOAuthLoginFailed = "oauth_login_failed"
 	ActionPlanChanged      = "plan_changed"
-	ActionAccountDeleted   = "account_deleted"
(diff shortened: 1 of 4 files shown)


xarmian:
@codex

PR #110, opened 2026-04-14
feat: hide plan settings in non-cloud mode, improve overrides UX

xarmian:
## Summary
- **Hide plan UI in non-cloud mode**: Plan column in user table, Plan dropdown, and Plan Overrides section are now gated behind `cloud_mode` so self-hosted admins don't see irrelevant billing controls
- **Replace JSON textarea with structured fields**: Plan overrides now use a labeled grid of number inputs (workspaces, items per workspace, members, API tokens, webhooks) instead of a raw JSON textarea — clearer, less error-prone, no JSON syntax to remember

## Test plan
- [ ] On a self-hosted instance: open Admin → Users — verify Plan column is hidden from the table and no plan/overrides fields appear in the user detail panel
- [ ] On a cloud instance: verify Plan column shows in the table, Plan dropdown and structured overrides grid appear in the detail panel
- [ ] Set an override (e.g. workspaces = 10), save — verify it persists on reload
- [ ] Clear all override fields, save — verify overrides are removed
- [ ] Enter -1 for a field — verify it saves as unlimited
- [ ] Enter non-numeric text — verify validation catches it before saving

Addresses IDEA-561

### DIFF
--- a/web/src/routes/console/admin/+page.svelte
+++ b/web/src/routes/console/admin/+page.svelte
@@ -15,7 +15,9 @@
 		{ key: 'api_tokens', label: 'API tokens', hint: 'Max API tokens' },
 		{ key: 'webhooks', label: 'Webhooks', hint: 'Max webhooks per workspace' },
 	];
+	const overrideFieldKeys = new Set(overrideFields.map(f => f.key));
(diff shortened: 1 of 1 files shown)


xarmian:
@codex

PR #111, opened 2026-04-14
fix: sidebar collection drag-and-drop only moving by one slot

xarmian:
## Summary
- Fixed sidebar collection drag-and-drop reordering only moving items by one position instead of to the drop target
- Root cause: `isDraggingSidebar` was set to `false` before API update calls, causing a Svelte `$effect` to reset `sidebarCollections` back to the store's original order after the first `await` yielded execution
- Fix: keep `isDraggingSidebar = true` during updates, use a local copy of the reordered array, fire all `sort_order` updates in parallel with `Promise.all`, and only release the flag after `loadCollections` completes

Fixes BUG-562

## Test plan
- [ ] Drag a collection in the sidebar from position 1 to position 5 — it should land at position 5
- [ ] Drag a collection up by multiple positions — verify correct placement
- [ ] Drag a collection by one position — still works as before
- [ ] Verify no visual "snap back" during the update

### DIFF
--- a/web/src/lib/components/layout/Sidebar.svelte
+++ b/web/src/lib/components/layout/Sidebar.svelte
@@ -64,6 +64,7 @@
 	// Drag-and-drop reordering state
 	let sidebarCollections: Collection[] = $state([]);
 	let isDraggingSidebar = $state(false);
+	let reorderGeneration = 0;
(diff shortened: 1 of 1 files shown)


xarmian:
@codex

PR #158, opened 2026-04-20
feat(web): add reusable BottomSheet primitive (TASK-627)

xarmian:
## Summary

Adds `$lib/components/common/BottomSheet.svelte` — a reusable bottom-sheet / modal primitive for the Pad web UI. Built on Svelte 5 runes with no new runtime dependencies.

**Why now:** First consumer is [[IDEA-493]] (the quick-actions menu gets clipped on mobile). Rather than bake a sheet into that one component, we're landing it as a standalone primitive we can reuse elsewhere (filter menus, mobile action sheets, etc.).

## Features

- **Mobile (< 640px):** docked to bottom, full-width, rounded top corners, swipe-down-to-dismiss (80px threshold via pointer events).
- **Desktop (>= 640px):** `desktopMode` prop — `'sheet'` (default, bottom-anchored with max-width) or `'centered'` (traditional centered dialog, mirrors existing `.overlay`/`.modal` pattern from `CreateCollectionModal.svelte`).
- **Dismissal:** Escape key, backdrop tap, or swipe-down all call `onclose()`.
- **Focus management:** focus trap while open; restores focus on close to the previously focused element.
- **Scroll lock:** body scroll locked while open, prior value restored on close/unmount.
- **Transitions:** `fly` for the sheet + `fade` for the backdrop; honors `prefers-reduced-motion`.
- **a11y:** `role="dialog"`, `aria-modal="true"`, optional `title` wired to `aria-labelledby`.

## API

```ts
interface Props {
  open: boolean;
  onclose: () => void;
  title?: string;
  fullHeight?: boolean;       // default false
  maxWidth?: string;          // desktop max-width
  desktopMode?: 'sheet' | 'centered';  // default 'sheet'
  children: Snippet;
}
```

Usage example lives in the doc-comment at the top of the component.

## Context

Implements `TASK-627` under `IDEA-493`. Parent for the other two tasks (`TASK-628`, `TASK-629`) which will consume this component for the quick-actions menu rework.

## Test plan

- [x] `go vet ./...` — clean
- [x] `go test ./...` — all tests pass
- [x] `cd web && npm run build` — clean (15.2s)
- [x] Svelte MCP `svelte-autofixer` — zero blocking issues
- [ ] Manual smoke: verify open/close/backdrop/escape/swipe in `make dev-web` on desktop + mobile viewport (Codex review + visual check on first consumer PR)

### DIFF
--- a/web/src/lib/components/common/BottomSheet.svelte
+++ b/web/src/lib/components/common/BottomSheet.svelte
@@ -188,21 +188,21 @@
 			}
 		});
 		return () => {
-			// Skip restoration if another sheet is still in control — otherwise
-			// closing a non-topmost stacked sheet could pull focus out of the
-			// still-active dialog and break modal isolation.
+			if (!previouslyFocused || typeof previouslyFocused.focus !== 'function') return;
+			if (!previouslyFocused.isConnected) return;
+
 			const othersStillOpen = openStack.some((t) => t !== myToken);
-			const activeDialog =
-				typeof document !== 'undefined' &&
-				document.activeElement instanceof HTMLElement
-					? document.activeElement.closest('[role="dialog"]')
-					: null;
-			const focusInOtherDialog = activeDialog !== null && activeDialog !== sheetEl;
-			if (othersStillOpen || focusInOtherDialog) return;
-			// Restore focus if the element is still connected.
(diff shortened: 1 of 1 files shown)


xarmian:
@codex review

PR #159, opened 2026-04-20
feat(web): mobile BottomSheet + viewport-aware dropdown in QuickActionsMenu (TASK-628)

xarmian:
## Summary

Fixes the mobile clipping bug in the quick-actions dropdown (Problem 1 from IDEA-493). When the trigger wrapped to the left edge of a narrow viewport, the 200px-min-width dropdown extended off-screen.

**Changes:**
- **Below 640px:** menu now renders as a `BottomSheet` (from TASK-627 / PR #158) — full-width, backdrop tap / Escape / swipe-down to dismiss.
- **Desktop popover** (>= 640px): kept, but hardened:
  - Viewport-aware alignment: flips from right-anchored to left-anchored when the trigger is within 220px of the viewport's left edge. Measured via `getBoundingClientRect()` at open time.
  - `max-width: calc(100vw - var(--space-4))` defensive overflow clamp.
- Action list shared between modes via a Svelte 5 snippet (no markup duplication).
- Outside-click handler short-circuits on mobile so the `BottomSheet` owns dismissal.

All existing behavior preserved: clipboard copy, toast on success/fail, trigger styling.

## Context

Implements `TASK-628` under `IDEA-493`. Consumer of the `BottomSheet` primitive landed in PR #158 (`TASK-627`).

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./...` — all tests pass
- [x] `cd web && npm run build` — clean
- [x] Svelte MCP `svelte-autofixer` — zero blocking issues
- [ ] Manual: verify mobile viewport (375px) opens bottom sheet, dismisses via all paths
- [ ] Manual: verify desktop near-left-edge trigger flips alignment correctly

PR #160, opened 2026-04-20
feat(web): inline + New and Manage affordances in QuickActionsMenu (TASK-629)

xarmian:
## Summary

Adds discovery paths for creating and editing quick actions directly from the menu, closing Problem 2 in IDEA-493. The editor already existed inside `EditCollectionModal` but was effectively invisible from the menu surface — users had to know to open collection settings, find the Quick Actions tab, and edit from there.

### `QuickActionsMenu.svelte`
Gated behind a new `canEdit` prop:
- **+ New quick action** footer row → toggles an inline form (icon picker + label + monospace prompt input + template-variable help line). On save, PATCHes the collection via `api.collections.update`, appends the new action to `settings.quick_actions`, and fires `oncollectionupdated` so the parent reloads. Toast on success / error.
- **⚙️ Manage actions** footer row → fires `onmanage`, which the parent wires to open `EditCollectionModal` deep-linked to the Quick Actions tab.
- Trigger button now stays visible for editors even when no actions exist yet, so they can bootstrap the first action without round-tripping through collection settings.

New props: `wsSlug: string`, `canEdit?: boolean`, `onmanage?: () => void`, `oncollectionupdated?: (c: Collection) => void`.

### `EditCollectionModal.svelte`
New `initialSection?: 'general' | 'fields' | 'display' | 'actions'` prop. When set, opens the modal directly to that tab instead of the default `'general'`. Default behavior unchanged when prop is omitted.

### Route wiring
- `[collection]/+page.svelte` — passes new props, tracks `editCollectionSection` for deep-linking.
- `[collection]/[slug]/+page.svelte` — same wiring, plus imports + renders `EditCollectionModal` inline (it wasn't previously rendered on item detail pages) so the "Manage actions" link works from item pages too.

## Context

Implements `TASK-629` under `IDEA-493`. Completes the three-task arc started in PR #158 (TASK-627 — BottomSheet primitive) and PR #159 (TASK-628 — mobile sheet + viewport-aware positioning).

## Test plan

- [x] `go build ./...` / `go vet ./...` / `go test ./...` — clean
- [x] `cd web && npm run build` — clean (15.2s)
- [x] Svelte MCP `svelte-autofixer` — zero blocking issues
- [ ] Manual: create a new quick action from the inline form on both collection list and item detail pages
- [ ] Manual: "Manage actions" opens the modal to the Quick Actions tab directly
- [ ] Manual: non-owners don't see the footer affordances
- [ ] Manual: works inside mobile bottom sheet and desktop popover

### DIFF
--- a/web/src/lib/components/collections/EditCollectionModal.svelte
+++ b/web/src/lib/components/collections/EditCollectionModal.svelte
@@ -26,11 +26,12 @@
 		open: boolean;
 		collection: Collection;
 		wsSlug: string;
+		initialSection?: 'general' | 'fields' | 'display' | 'actions';
 		onupdated: (updated?: Collection) => void;
 		onclose: () => void;
 	}
 
-	let { open, collection, wsSlug, onupdated, onclose }: Props = $props();
+	let { open, collection, wsSlug, initialSection, onupdated, onclose }: Props = $props();
 
 	let confirmArchive = $state(false);
 	let archiving = $state(false);
@@ -193,7 +194,7 @@
 			}));
 			newFields = [];
 			error = '';
-			activeTab = 'general';
+			activeTab = initialSection ?? 'general';
 			confirmArchive = false;
 
 			// Sync display settings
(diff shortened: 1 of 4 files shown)


xarmian:
@codex review

PR #170, opened 2026-04-21
fix(web): sanitize rendered markdown through DOMPurify (TASK-647)

xarmian:
## Summary
Pipe `renderMarkdown` output through `DOMPurify.sanitize` with a strict markdown-only allowlist. Closes the stored XSS vector where a comment could inject `<script>` / inline event handlers / javascript: URLs that ran on every viewer's page.

## Context
Implements `TASK-647` (C1) under `PLAN-643` (OSS Security Hardening). The task called out `web/src/lib/utils/markdown.ts:43` returning raw `marked()` output, with `CommentThread.svelte:151` and `TimelineCommentCard.svelte:152,223` piping it straight to `{@html}`.

## Changes
- Added `sanitizeMarkdownHtml(html)` helper using DOMPurify with an explicit `ALLOWED_TAGS` / `ALLOWED_ATTR` allowlist covering exactly what marked + our custom wiki-link renderer produce.
- `renderMarkdown` now sanitizes before returning — all call sites (`CommentThread`, `TimelineCommentCard`, and any future consumer) are safe by default.
- HTML-escape the wiki-link title in the intermediate HTML so attribute boundaries can't be broken, keeping pre-sanitization markup well-formed.
- In SSR/prerender (no `window`) the sanitizer returns `""` — user-generated markdown is never prerendered in adapter-static SPA mode, so this is a safe no-op.

## Test plan
- [x] `go vet ./...` — clean
- [x] `go test ./...` — all pass
- [x] `cd web && npm run build` — clean
- [x] Existing svelte-check errors in unrelated routes are pre-existing (not introduced by this change)

### DIFF
--- a/web/src/lib/utils/markdown.ts
+++ b/web/src/lib/utils/markdown.ts
@@ -30,7 +30,9 @@ const MARKDOWN_ALLOWED_TAGS = [
 // strips javascript: and data: hrefs via its default URL policy.
 const MARKDOWN_ALLOWED_ATTR = [
 	'href', 'title', 'target', 'rel', 'class', 'aria-hidden',
-	'alt', 'src', 'id', 'name', 'align', 'type', 'checked', 'disabled'
+	'alt', 'src', 'id', 'name', 'align', 'type', 'checked', 'disabled',
+	// <ol start="N"> — marked emits this for lists that don't begin at 1.
+	'start'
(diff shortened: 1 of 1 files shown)


xarmian:
@codex review

xarmian:
@codex review

PR #172, opened 2026-04-21
feat(server): strict-dynamic CSP + fail-fast missing index.html (TASK-375)

xarmian:
## Summary
Completes the remaining items on the nonce-based CSP work: add \`'strict-dynamic'\` to \`script-src\`, and fail fast when the embedded \`index.html\` can't be read.

## Context
Implements \`TASK-375\` under \`PLAN-643\` (OSS Security Hardening). The per-request nonce, single-read of \`index.html\`, and \`script-src 'nonce-…'\` (no \`unsafe-inline\`) were already in place; this PR adds the two outstanding items from the task's implementation checklist.

## Changes
- \`internal/server/server.go\` — add \`'strict-dynamic'\` to script-src. In CSP-L3 browsers this supersedes the \`'self'\` host-list for scripts, so a future XSS that injects \`<script src="//evil">\` is blocked even though \`'self'\` is present (kept as a fallback for older browsers). SvelteKit's bootstrap script already uses dynamic \`import()\` for runtime chunks, which is exactly what strict-dynamic permits.
- Fail fast on \`index.html\` read error — previous silent-swallow returned blank HTML forever. Panic at startup so a broken build is obvious.

## Test plan
- [x] \`go test ./internal/server/...\` — passes

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -537,7 +537,13 @@ func (s *Server) SetWebUI(fsys fs.FS) {
 
 func (s *Server) spaHandler() http.Handler {
 	fileServer := http.FileServer(http.FS(s.webFS))
-	indexHTML, _ := fs.ReadFile(s.webFS, "index.html")
+	indexHTML, err := fs.ReadFile(s.webFS, "index.html")
+	if err != nil {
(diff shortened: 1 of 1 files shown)


xarmian:
@codex review

PR #180, opened 2026-04-22
feat(server): gate /metrics behind loopback + bearer token (TASK-653)

xarmian:
## Summary
\`/metrics\` is no longer exposed to every caller on the network. Default: loopback-only. With \`PAD_METRICS_TOKEN\` set: \`Authorization: Bearer <token>\` required (constant-time compared).

## Context
Implements \`TASK-653\` (H4) under \`PLAN-643\` (OSS Security Hardening). Previously \`/metrics\` was served outside the auth middleware group — unauthenticated scrapes revealed workspace counts, API usage patterns, and (via labels) user/workspace IDs.

## Changes
- \`internal/server/server.go\` — new \`metricsAuth\` middleware wraps the \`/metrics\` group. \`subtle.ConstantTimeCompare\` on the token. \`WWW-Authenticate: Bearer realm="metrics"\` on 401s.
- \`SetMetricsToken(string)\` server API.
- \`internal/config/config.go\` — \`MetricsToken\` + \`PAD_METRICS_TOKEN\` env.
- \`cmd/pad/main.go\` — plumb config to server.
- \`.env.example\` — document \`PAD_METRICS_TOKEN\` with \`openssl rand -hex 32\` hint.
- \`internal/server/metrics_auth_test.go\` — coverage: loopback-allowed/LAN-denied when token unset; missing/wrong/correct Bearer when set; non-Bearer scheme rejected; \`WWW-Authenticate\` present on 401; \`/metrics\` returns 404 when \`SetMetrics\` wasn't called.

## Test plan
- [x] \`go test ./...\` — all pass (3 new test cases)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -34,3 +34,10 @@ POSTGRES_PASSWORD=
 # OPTIONAL — Trusted reverse-proxy CIDRs. Only peers in these CIDRs may set
 # X-Forwarded-For / X-Real-IP. Leave unset for direct-exposed deploys.
 # PAD_TRUSTED_PROXIES=10.0.0.0/8,172.16.0.0/12
+
+# OPTIONAL — Bearer token required to scrape /metrics. When unset, /metrics
+# is reachable only from loopback (so a Prometheus on the same host keeps
+# working without config); when set, every scrape must send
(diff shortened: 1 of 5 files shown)


xarmian:
@codex review

PR #188, opened 2026-04-22
fix(server): reject CORS wildcard when credentials are on (TASK-664)

xarmian:
## Summary
Drop \`*\` from \`PAD_CORS_ORIGINS\` (it's incompatible with \`AllowCredentials=true\`) and default \`AllowCredentials\` to false when no origins are explicitly configured.

## Context
Implements \`TASK-664\` (L1) under \`PLAN-643\` (OSS Security Hardening). Previously a typo like \`PAD_CORS_ORIGINS=*\` looked fine in curl but failed silently from every real browser, and empty \`PAD_CORS_ORIGINS\` ran with credential sharing on — letting a browser on a different origin piggy-back cookies on the victim's session.

## Changes
- \`internal/server/middleware_security.go\`:
  - \`parseCORSOrigins\` — explicit \`*\` dropped with a log warning; lone \`*\` falls back to localhost defaults.
  - New \`corsAllowCredentials(corsOrigins)\` — returns true only when an operator has set \`PAD_CORS_ORIGINS\`.
- \`internal/server/server.go\` — \`cors.Options.AllowCredentials = corsAllowCredentials(s.corsOrigins)\`.
- \`internal/server/middleware_security_test.go\` — coverage for \`*\` handling (lone / mixed / trailing) and \`corsAllowCredentials\` (empty / whitespace / explicit).

## Test plan
- [x] \`go test ./...\` — all pass

### DIFF
--- a/internal/server/middleware_security.go
+++ b/internal/server/middleware_security.go
@@ -3,6 +3,7 @@ package server
 import (
 	"crypto/rand"
 	"encoding/base64"
+	"log/slog"
 	"net/http"
 	"strings"
 )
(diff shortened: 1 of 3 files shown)


xarmian:
@codex review

PR #189, opened 2026-04-22
fix(config): auto-generate PAD_ENCRYPTION_KEY on first run (TASK-668)

xarmian:
## Summary
Encryption is now mandatory. If no \`PAD_ENCRYPTION_KEY\` is provided, Pad persists a freshly-generated 32-byte AES-256 key to \`<DataDir>/encryption.key\` with 0600 permissions on first run, and reuses it across restarts.

## Context
Implements \`TASK-668\` (L5) under \`PLAN-643\` (OSS Security Hardening). Previously an empty key caused silent plaintext storage of sensitive fields (TOTP seeds) with only a startup WARN.

## Changes
- \`internal/config/config.go\`:
  - New \`EncryptionKeyFile()\` + \`EnsureEncryptionKey()\`. Resolves env → config → file → generate, in that order.
  - \`EncryptionKeySource\` string field surfaces which path was taken so callers can log loudly.
- \`cmd/pad/main.go\`:
  - Drop the "empty-key → warn → continue-plaintext" branch.
  - Call \`cfg.EnsureEncryptionKey()\`; fail startup on error; log at WARN when the key was newly generated so operators notice the new file.
- \`internal/config/encryption_key_test.go\`: table tests — generates when missing (file 0600, 32 raw bytes); loads from file; respects configured key; idempotent across restarts.

## Test plan
- [x] \`go test ./...\` — all pass (4 new tests)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,6 +11,14 @@
 # REQUIRED — PostgreSQL password for the `pad` user.
 POSTGRES_PASSWORD=
 
+# REQUIRED on Postgres — AES-256 encryption key for TOTP seeds and other
+# sensitive fields. 64-character hex string (32 bytes). Pad will auto-
+# generate one on SQLite deployments but refuses to on Postgres because
+# multi-replica deployments would generate divergent keys and fail cross-
+# instance decryption. Generate with:
+#   openssl rand -hex 32
+PAD_ENCRYPTION_KEY=
+
 # OPTIONAL — Which host interface the Pad web UI publishes to.
 # Default (loopback only) is the safest choice on a fresh install, because
 # the bootstrap endpoint is reachable until the first admin is created.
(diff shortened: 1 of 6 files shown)


xarmian:
@codex review

xarmian:
@codex review

xarmian:
@codex review — latest commit 7720ba1 uses O_CREATE|O_EXCL with reload-on-EEXIST (TestEnsureEncryptionKey_ConcurrentStartIsRaceSafe passes under -race). Please re-evaluate.

xarmian:
@codex review

xarmian:
@codex review

xarmian:
@codex review — latest commit b5c3160 removes the startup-blocking gate (EnsureEncryptionKey(true) always) and warns only when generating on Postgres. Please re-evaluate.

xarmian:
@codex review

PR #190, opened 2026-04-22
fix(server): derive CLI auth URL scheme from r.TLS, gate X-Forwarded-Proto on trusted proxies (TASK-665)

xarmian:
## Summary
`handleCreateCLIAuthSession` accepted `X-Forwarded-Proto` from any client to choose the scheme for the browser auth URL printed by `pad auth login`. On a plain-HTTP self-host deployment an attacker could forge `https://` in that link — low-impact phishing (user clicks in their own terminal), but the safe default is to ignore unauthenticated proxy headers.

New precedence in `cliAuthScheme`:
1. `r.TLS != nil` → `https`
2. Peer in `PAD_TRUSTED_PROXIES` → honor `X-Forwarded-Proto` (first comma-separated value, case-insensitive, must be `http`/`https`)
3. Otherwise → `http`

Uses `rawPeerAddr` so the check still works after `TrustedProxyRealIP` rewrites `r.RemoteAddr`.

## Context
Implements `TASK-665` under `PLAN-643` (OSS Security Hardening).

## Test plan
- [x] `go vet ./...` — clean
- [x] `go test ./...` — all tests pass
- [x] New table-driven test `TestCLIAuthScheme` covers TLS, untrusted-peer spoofing, trusted-peer forwarding, chained/case-insensitive/garbage `X-Forwarded-Proto` values

### DIFF
--- a/internal/server/handlers_cli_auth.go
+++ b/internal/server/handlers_cli_auth.go
@@ -2,7 +2,9 @@ package server
 
 import (
 	"fmt"
+	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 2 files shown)


xarmian:
@codex review

PR #192, opened 2026-04-22
fix(server): deny-by-default whitelist for API token scopes (TASK-667)

xarmian:
## Summary
\`tokenScopeAllows\` previously fell open on unrecognized scopes AND on unparseable scope JSON. A typo like \`\"read-only\"\` silently granted full access — exactly the kind of landmine that a fresh token minted by an admin who misremembers the vocabulary would step on.

**New policy (deny-by-default):**
- Unparseable JSON → **deny** + warn (was allow). Data corruption / tampering never falls open.
- Unrecognized scopes → never contribute to allow; all unknowns on a request log once so operators can spot typos.
- \`\"*\"\` and \`\"write\"\` still allow all methods; \`\"read\"\` still allows safe methods only.
- Empty string and empty array \`[]\` still allow — legacy pre-enforcement rows shouldn't break on upgrade.

## Context
Implements \`TASK-667\` under \`PLAN-643\` (OSS Security Hardening).

## Test plan
- [x] \`go test ./...\` — all tests pass
- [x] Updated test table: \"unknown scope allows GET/POST\" → deny; \"read-only typo denies GET\" regression pin; \"unknown+wildcard still allows\" guard rails; \"invalid json denies all\" (was allow).

### DIFF
--- a/internal/server/middleware_auth.go
+++ b/internal/server/middleware_auth.go
@@ -690,14 +690,9 @@ func clearSessionCookie(w http.ResponseWriter, secure bool) {
 func tokenScopeAllows(scopesJSON, method, path string) bool {
 	_ = path // reserved for future per-resource scopes
 
-	// Legacy tokens with no scopes column → full access. Same fast path
-	// for the explicit wildcard and for the explicit empty array (legacy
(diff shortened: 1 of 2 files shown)


xarmian:
@codex review

PR #194, opened 2026-04-22
feat(server): periodically revalidate SSE subscriber membership (TASK-670)

xarmian:
## Summary
\`handleSSE\` verified workspace access only at connection time. A removed member kept receiving events until they manually disconnected — or, more commonly, indefinitely because \`EventSource\` auto-reconnects. An owner who revoked access had no way to stop the leak.

- New 60s membership revalidation ticker inside the SSE select loop.
- \`sseSubscriberStillHasAccess\` mirrors \`RequireWorkspaceAccess\`'s matrix: fresh-install bypass, admin role, direct membership, guest grants, legacy workspace-scoped API token. DB errors fail OPEN (transient blip doesn't bounce every open tab); missing membership fails CLOSED.
- On revocation we emit \`{type:\"unauthorized\"}\` with a reason BEFORE closing the stream so the frontend can route to login instead of tight-loop reconnect.
- \`sseMembershipRevalInterval\` is a package-level var for test injection; pinned to a sensible 30-300s range.

## Context
Implements \`TASK-670\` under \`PLAN-643\` (OSS Security Hardening).

## Test plan
- [x] \`go test ./...\` — all tests pass
- [x] Unit test covers admin, active member, outsider, removed member, guest-grant (skipped if default collections aren't seeded), unauthenticated, legacy token scoped to same workspace, and legacy token scoped to a different workspace.

### DIFF
--- a/internal/server/handlers_events.go
+++ b/internal/server/handlers_events.go
@@ -88,88 +88,19 @@ func (s *Server) handleSSE(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// Compute the user's visible collection set for event filtering.
-	// Build a slug-based set since events carry collection slugs, not IDs.
-	var visibleSlugSet map[string]bool // nil = all access (no filtering)
-	visibleIDs, err := s.visibleCollectionIDs(r, ws.ID)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


xarmian:
@codex review

PR #219, opened 2026-04-23
chore: run containers as non-root + harden K8s securityContext (TASK-680)

xarmian:
## Summary
Both Dockerfiles now run as an unprivileged `uid:1000` user, and the K8s deployment gets a full hardened securityContext. First thing external reviewers flag on a public OSS repo.

## Context
Implements `TASK-680` under `PLAN-644` (OSS Repo Hygiene & Launch Polish).

### Dockerfile changes (`Dockerfile` + `Dockerfile.goreleaser`)
- Add non-login `uid:1000` user `pad` via `adduser -D`
- `chown -R pad:pad /data` so SQLite writes from the unprivileged user
- `USER pad` before `ENTRYPOINT`

### K8s changes (`deploy/k8s/deployment.yaml`)
**Pod-level `securityContext`:**
- `runAsNonRoot: true`
- `runAsUser: 1000`, `runAsGroup: 1000`, `fsGroup: 1000` (fsGroup makes the mounted emptyDir writable by the group)
- `seccompProfile.type: RuntimeDefault`

**Container-level `securityContext`:**
- `allowPrivilegeEscalation: false`
- `readOnlyRootFilesystem: true`
- `capabilities.drop: [ALL]`

### Why read-only rootfs is safe
The server writes only to `/data` (SQLite DB + encryption key). `internal/cli/editor.go` uses `os.CreateTemp` but that's CLI-only — `pad server start` never touches `/tmp`. Verified by running the container with `--read-only --tmpfs /tmp:rw` and confirming the server boots, serves /api/v1/health, and logs no filesystem errors.

## Test plan
- [x] `docker build -t pad .` succeeds
- [x] `docker inspect pad --format '{{.Config.User}}'` returns `pad`
- [x] `docker run pad` + `docker exec pad id` returns `uid=1000(pad) gid=1000(pad)`
- [x] `docker run --read-only --tmpfs /tmp:rw pad` starts cleanly; `/api/v1/health` returns success
- [x] `deploy/k8s/deployment.yaml` parses via `yq`; `.spec.template.spec.securityContext` + `.spec.template.spec.containers[0].securityContext` blocks render correctly
- [x] `go build ./... && go vet ./... && go test ./...` green

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,11 +21,14 @@ FROM alpine:3.21
 RUN apk add --no-cache ca-certificates tzdata
 COPY --from=go-builder /app/pad /usr/local/bin/pad
 
-# Data directory
-RUN mkdir -p /data
+# Create a non-root user (uid 1000) and data directory owned by it
+RUN adduser -D -u 1000 -h /home/pad pad \
+    && mkdir -p /data \
+    && chown -R pad:pad /data
 ENV PAD_DATA_DIR=/data
 ENV PAD_HOST=0.0.0.0
 
+USER pad
 EXPOSE 7777
 VOLUME /data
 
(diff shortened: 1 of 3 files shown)


xarmian:
@codex review

PR #242, opened 2026-04-25
fix(web): wire desktop topbar workspace switching through last-route restore

xarmian:
Follow-up to PR #240. Restore worked only via the mobile WorkspaceSwitcher; desktop topbar used plain `<a href>` to dashboard, bypassing restore. Now intercepted on left-click via shared helper. Modifier-clicks / middle-click still go to dashboard via the original href.


xarmian:
Closing — over-engineered. Will land simpler.

PR #565, opened 2026-05-16
feat(store): wrap migrations in atomic transactions (IDEA-1485)

xarmian:
## Summary

Wrap each migration file's execution in a transaction so a mid-migration crash (process kill, OOM, power loss) cannot leave the DB in a partially-mutated state with the bookkeeping row absent — the failure shape that risks permanent data loss for the table-rebuild migration pattern used in `022_audit_trail.sql` and `055_collections_settings_not_null.sql`.

Closes IDEA-1485.

## Plus

- **PRAGMA-aware tx wrapping.** SQLite's `PRAGMA foreign_keys` is per-connection AND a no-op inside a transaction. The runner pins each migration to one pool connection via `db.Conn(ctx)`, then detects PRAGMA statements in the migration text and emits them around (not inside) the wrapping tx:
  - `PRAGMA foreign_keys = OFF/0/false` → executed before `BEGIN`
  - `PRAGMA foreign_keys = ON/1/true` → executed after `COMMIT`
  - Other PRAGMAs (e.g. `journal_mode = WAL` in 001) → executed after `COMMIT` defensively
- **FK restore on all exit paths.** If a migration disables FKs and crashes before re-enabling them, a deferred `PRAGMA foreign_keys = ON` runs before the pinned conn returns to the pool — so the conn doesn't leak `OFF` for subsequent pool checkouts. Verified by a regression test that fails when the defer is removed.
- **Postgres path mirrors the shape** — no PRAGMA detection (no PRAGMAs in Postgres), but the same explicit `BeginTx` / `Commit` / `Rollback`-on-error wrap. Multi-statement migration text is exec'd as one `tx.Exec()` call; Postgres handles that atomically inside a tx. The corpus contains no `CREATE INDEX CONCURRENTLY` or other non-tx-safe statements.
- **`execMulti` generalized** from `*sql.DB` to a local `sqlExecer` interface so it can take a `*sql.Tx` for the wrapping. Only caller is the new `applySQLiteMigration`.

## Behavior changes

- Migration files now apply atomically: the `schema_migrations` row is INSERT'd inside the same transaction as the migration's DDL. On crash, either the whole migration committed or none of it did — the previous "DDL partial, bookkeeping absent, re-run on restart drops the half-rebuilt table" failure mode is eliminated.
- For SQLite migrations that toggle `foreign_keys`: the toggle now actually takes effect (it was a no-op inside the wrapping tx if not lifted). For pre-existing migrations 022 and 055, behavior is functionally unchanged because they already needed FK-off and the lifting preserves that intent.
- No public API or schema changes.

## Test plan

- [x] SQLite: `go test ./...` passes (store suite ~42s).
- [x] Postgres: `PAD_TEST_POSTGRES_URL=<url> go test ./...` passes (store suite ~101s).
- [x] New regression tests:
  - `TestMigrationAtomicity_FailedSQLite_RollsBackPartialDDLAndBookkeeping`
  - `TestMigrationAtomicity_FailedPostgres_RollsBackPartialDDLAndBookkeeping` (skipped without PG URL)
  - `TestMigrationAtomicity_PragmaForeignKeysLifted` (asserts FK-rebuild migration succeeds and FK enforcement restored)
  - `TestMigrationAtomicity_FailedSQLite_RestoresForeignKeysOnError` (asserts deferred FK restore covers error paths)
  - `TestMigrationAtomicity_SuccessfulSQLite_RecordsBookkeeping` (positive control)
  - `TestExtractPragmas` (unit test for the PRAGMA line-scan extractor)
- [x] Defer correctness verified by removing the defer and observing the test failure.

## Refs

- IDEA-1485 (this work)
- IDEA-1484 / PR #562 / #563 (the migration shape that surfaced this gap during R1 review)
- BUG-1482 / PR #561 (the precedent migration that introduced the table-rebuild pattern)
- DECIS-32 in `claude` workspace — orchestration data-point-8

### DIFF
--- a/internal/store/migration_atomicity_test.go
+++ b/internal/store/migration_atomicity_test.go
@@ -1,6 +1,7 @@
 package store
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


xarmian:
Web check failure is the pre-existing npm audit flake (devalue/mermaid/svelte CVEs in node_modules) — unrelated to this PR which only touches Go store code. Per PLAYB-40 convention (HANDO-48 documents the 6-day streak). Go and Go (PostgreSQL) both pass; that's the dual-driver matrix this change targets. Proceeding with squash-merge.

PR #669, opened 2026-05-30
feat(api): bulk-mutation endpoint + single SSE batch event (TASK-1668)

xarmian:
## Summary
Adds `POST /api/v1/workspaces/{ws}/items/bulk` — a single endpoint that applies one mutation verb to many items, emitting **one** SSE batch event and **one** webhook for the whole batch instead of per-item fan-out.

Verbs: `archive`, `move` (status and/or collection), `tag`, `untag`, `set-priority`, `assign`.

The lane-header bulk actions (PLAN-1667) operate on a whole filtered lane, so per-item SSE/webhook events would flood subscribers. This endpoint is the backend half — frontend client + TS types are TASK-1669, UI wiring is TASK-1672.

## Design
- **Reuses store paths** — `UpdateItemWithPreCheck` / `MoveItem` / `DeleteItem`; no new write logic.
- **Open-children guard** runs per status-bearing move exactly as the single PATCH path does (force-overridable via `force: true`).
- **Per-row failures collected**, not fatal — response envelope is `{ op, updated:[{ref,id}], failed:[{ref,error,code?,details?}], total }`. `open_children` rejections carry structured `code` + `details`.
- **Editor/owner gated** (`canEdit` in the UI) — viewers/guests get 403.
- **One SSE event**: `items_bulk_updated` `{ op, item_ids, count, seq:<max>, actor }`. New event type + `ItemIDs`/`Op`/`Count` fields on `events.Event`. Emitted with `event: items_bulk_updated` via the existing `writeSSEEvent` path.
- **One webhook**: `item.bulk_updated` `{ op, count, item_ids }`.
- Route registered as a static `/items/bulk` segment **before** `/items/{itemSlug}` so it isn't captured as a slug.
- Batch capped at 1000 items as a guardrail.

## Wire contract (for TASK-1669)
```
op=archive       — no params
op=move          — status and/or collection (collection = target slug)
op=set-priority  — priority
op=tag | untag   — tags: []
op=assign        — assigned_user_id?, agent_role_id?, clear_assigned_user?, clear_agent_role?
```

## Test plan
- [x] go build ./... — clean
- [x] go vet ./... — clean
- [x] go test ./... — all pass (new `handlers_items_bulk_test.go`: set-priority, move-status, tag/untag idempotence, archive, partial failure, validation matrix, route registration)
- [x] No `web/` changes — backend only

### DIFF
--- a/internal/events/bus.go
+++ b/internal/events/bus.go
@@ -76,16 +76,20 @@ type Event struct {
 	// comment_*, reaction_*) and for legacy publishers that
 	// haven't been upgraded.
 	Seq int64 `json:"seq,omitempty"`
-	// ItemIDs / Op / Count describe a batch event (ItemsBulkUpdated,
-	// TASK-1668). ItemIDs lists the affected item IDs; Op is the verb
-	// applied (archive/move/tag/untag/set-priority/assign); Count is
-	// len(ItemIDs). Empty/zero for single-item events. A batch event
-	// carries no per-item seq — Seq holds the max seq across the batch
-	// so a local-first client can decide whether to apply deltas or
-	// trigger a /items-changes backfill.
-	ItemIDs []string `json:"item_ids,omitempty"`
-	Op      string   `json:"op,omitempty"`
-	Count   int      `json:"count,omitempty"`
+	// Op / Count describe a batch event (ItemsBulkUpdated, TASK-1668).
+	// Op is the verb applied (archive/move/tag/untag/set-priority/
+	// assign); Count is the number of items affected in this event's
+	// Collection. Zero/empty for single-item events.
+	//
+	// A batch event is scoped to ONE Collection (the bulk endpoint emits
+	// one per affected collection) so the SSE visibility filter routes
+	// it like any collection-scoped event. It deliberately carries NO
+	// per-item IDs: a batch can't be item-grant-filtered for guests on a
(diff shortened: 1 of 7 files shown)


xarmian:
## Codex review status (paused at safety cap)

Ran 7 Codex review rounds; rounds 1–6 each surfaced a distinct, real issue in this diff, all fixed:

1. Per-item collection visibility — restricted editors could bulk-mutate hidden-collection items by ref-guessing.
2. Frontend consumer — `items_bulk_updated` had no listener; other tabs stayed stale.
3. SSE routing leak — empty `Collection` leaked batch metadata to restricted members + starved guests.
4. Source-scope notify — collection moves only notified the target lane.
5. Item-grant-only leak — itemless batch events reached item-grant-only subscribers.
6. Status-override validation — collection-move status override bypassed target-schema validation.

**Round 7** flagged a *pre-existing, out-of-scope* limitation: `/items-changes` filters deltas by an item's **current** collection visibility, so an item moving from a visible→invisible collection returns no moved-out tombstone and lingers in a restricted member's local cache until rebootstrap. This affects single-item moves identically and predates this PR (which actually *added* the source-collection reconcile signal). Filed as **BUG-1675** (severity low; fix requires changing the shared delta endpoint to emit moved-out tombstones).

Paused before merge at the review-loop safety cap given the security-sensitive visibility routing — awaiting maintainer merge decision.

PR #732, opened 2026-06-15
Fix SQLite NOT NULL on api_tokens.workspace_id (account-token 500)

### LINKED ISSUE #731: SQLite: account-settings API token creation 500s — NOT NULL constraint on api_tokens.workspace_id
## Summary
Creating a **personal/account API token** from the settings page fails on **SQLite** deployments with a 500 (\"An internal error occurred\"). Postgres is unaffected.

## Symptom
Server log (pad 0.7.0, SQLite backend):
```
level=ERROR msg="internal server error" error="insert api token: constraint failed: NOT NULL constraint failed: api_tokens.workspace_id (1299)"
level=ERROR msg="http request" method=POST path=/api/v1/auth/tokens status=500
```

## Root cause
`handleCreateUserToken` (`internal/server/handlers_tokens.go`) intentionally creates a workspace-agnostic token — it never sets `input.WorkspaceID`. `Store.CreateAPIToken` (`internal/store/api_tokens.go`) then inserts `NULL` into `workspace_id`.

But the SQLite schema declares the column `NOT NULL`:
- `internal/store/migrations/011_api_tokens.sql` → `workspace_id TEXT NOT NULL`

…which diverges from both:
- **Postgres** — `internal/store/pgmigrations/001_initial.sql` → `workspace_id TEXT REFERENCES workspaces(id)` (nullable), and
- **the Go layer**, which treats it as nullable everywhere (`getAPIToken` / `ValidateToken` scan it into `*string`).

So the account-token path fails on every SQLite deployment. It's untested: all cases in `internal/store/api_tokens_test.go` pass a `WorkspaceID`, and there's no handler test for `POST /api/v1/auth/tokens`.

## Impact
Self-hosted SQLite users can't create API tokens from the account settings page (the per-workspace `handleCreateToken` path sets a workspace and works). This blocks generating tokens for external integrations.

## Fix
Rebuild `api_tokens` on SQLite with `workspace_id` nullable, matching Postgres and the code contract. PR incoming.

## Workaround
Insert the token row directly with a `workspace_id` set (sidesteps the constraint); the validate path accepts it.

b4rk13:
Fixes #731.

## Problem
On **SQLite** deployments, creating a personal/account API token from the settings page (`POST /api/v1/auth/tokens` → `handleCreateUserToken`) returns a 500:

```
error="insert api token: constraint failed: NOT NULL constraint failed: api_tokens.workspace_id (1299)"
```

`handleCreateUserToken` creates a workspace-agnostic token and never sets `WorkspaceID`, so `Store.CreateAPIToken` inserts `NULL` into `workspace_id`. But `migrations/011_api_tokens.sql` declared that column `NOT NULL` — diverging from:
- **Postgres**: `pgmigrations/001_initial.sql` declares `workspace_id TEXT REFERENCES workspaces(id)` (nullable), and
- **the Go layer**: `getAPIToken` / `ValidateToken` already scan `workspace_id` into `*string`.

Postgres is therefore unaffected; this is SQLite-only. The path was untested — every case in `api_tokens_test.go` passes a `WorkspaceID`.

## Fix
- **`migrations/068_api_tokens_workspace_nullable.sql`** — rebuild `api_tokens` with `workspace_id` nullable. SQLite has no `ALTER COLUMN ... DROP NOT NULL`, so it uses the established table-rebuild recipe (cf. `055_collections_settings_not_null.sql`), including the `PRAGMA foreign_keys = OFF/ON` toggle the migration runner explicitly supports. No table FK-references `api_tokens(id)`, so there are no inbound references to recreate. No Go changes needed.
- **`api_tokens_test.go`** — `TestCreateAPITokenUserScoped`: regression test for the workspace-agnostic path (create + validate with no workspace).

## Verification
Reproduced the exact failure and confirmed the fix against SQLite 3.50.4: pre-migration insert of a NULL-workspace token fails with the reported error; post-migration the existing scoped rows are preserved, the unscoped insert succeeds (`workspace_id` = NULL), and `name` remains `NOT NULL`.

> Note: I couldn't run the full Go suite in my environment (no Go toolchain), but the new test only uses existing helpers (`testStore`, `createTestUser`) and mirrors the surrounding cases.

### DIFF
--- a/internal/store/api_tokens_test.go
+++ b/internal/store/api_tokens_test.go
@@ -6,6 +6,41 @@ import (
 	"github.com/PerpetualSoftware/pad/internal/models"
 )
 
+// TestCreateAPITokenUserScoped covers the account-settings path
+// (handleCreateUserToken), which creates a workspace-agnostic token with no
+// WorkspaceID set. The store inserts NULL for workspace_id; this must succeed
+// and the token must validate. Regression test for the SQLite-only NOT NULL
+// constraint on api_tokens.workspace_id fixed in migration
+// 068_api_tokens_workspace_nullable.sql (Postgres already allowed NULL).
+func TestCreateAPITokenUserScoped(t *testing.T) {
+	s := testStore(t)
+	u := createTestUser(t, s, "test@test.com", "Test", "password123")
+
+	token, err := s.CreateAPIToken(u.ID, models.APITokenCreate{
+		Name: "user-scoped",
+	}, 90, 0)
+	if err != nil {
+		t.Fatalf("CreateAPIToken (no workspace) error: %v", err)
+	}
+	if token.Token == "" {
+		t.Fatal("expected plaintext token")
+	}
+	if token.WorkspaceID != "" {
+		t.Errorf("expected empty workspace, got %q", token.WorkspaceID)
+	}
+
+	validated, err := s.ValidateToken(token.Token)
+	if err != nil {
+		t.Fatalf("ValidateToken error: %v", err)
+	}
+	if validated == nil {
+		t.Fatal("expected user-scoped token to validate")
+	}
+	if validated.WorkspaceID != "" {
+		t.Errorf("expected empty workspace on validate, got %q", validated.WorkspaceID)
+	}
+}
+
 func TestCreateAPITokenWithExpiry(t *testing.T) {
 	s := testStore(t)
 	u := createTestUser(t, s, "test@test.com", "Test", "password123")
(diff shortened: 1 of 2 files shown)


xarmian:
Thanks for the outstanding report and fix, @b4rk13! 🙏

This is a model PR — spot-on root-cause analysis of the SQLite↔Postgres `workspace_id` divergence, a fix that follows the repo's established table-rebuild recipe (matching `055_collections_settings_not_null.sql`, including the lifted-PRAGMA `foreign_keys` handling), and a clean regression test for the unscoped-token path.

We checked out the branch and ran the full store suite locally: `TestCreateAPITokenUserScoped` passes, the entire migration chain (001→068) applies cleanly on a fresh DB, and `go build ./...` is green — so the one thing you couldn't verify without a Go toolchain is confirmed on our end.

Merging now; it'll ship in the upcoming **0.8.0** release. Thanks again for making Pad better for self-hosted SQLite users! 🎉

Issue #850, opened 2026-07-08
CLI: unknown subcommands should exit non-zero


xarmian:
### Problem
Running an unknown subcommand (a typo like `pad iem list`, or a removed command) prints the parent command's help and exits **0**. For an agent-first CLI that's a silent failure — typos and dead commands look like success, and scripts/agents can't detect them.

### Fix
Make parent commands that have subcommands return a **non-zero** exit on an unrecognized subcommand, instead of falling through to help + exit 0. With Cobra this is usually a fallback `RunE` (or `Args` validation) on the parent that returns an "unknown command" error.

### Pointers
- Commands live in `cmd/pad/main.go` (Cobra).
- Add a test asserting a bogus subcommand exits non-zero.

Small, self-contained, testable.

PR #861, opened 2026-07-08
fix: parent commands exit non-zero on unknown subcommand

molloyzak13:
Fixes PICK: PerpetualSoftware/pad | #850 | Go | CLI: unknown subcommands should exit non-zero.

Note: PR body was not generated by the contributor agent. See outbox/pr-PerpetualSoftware-pad-20260708-022626.md for the full plan.

### DIFF
--- a/cmd/pad/cmd_attachment.go
+++ b/cmd/pad/cmd_attachment.go
@@ -27,6 +27,7 @@ func attachmentCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "attachment",
 		Short: "Upload, download, view, and list item attachments",
+		RunE:  unknownSubcommandRun,
 		Long: `Upload, download, view, and list attachments (images, files) on items.
 
 Examples:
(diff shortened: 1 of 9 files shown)


xarmian:
Merged — thank you, @molloyzak13! 🙏

Your fix was clean and well-targeted: a single `unknownSubcommandRun` helper wired into the command groups, plus a focused test. Exactly the right shape for the problem, and it read clearly.

One thing I added before merging: the codebase recently split `cmd/pad/main.go` into per-resource files (`cmd_*.go`), and a handful of groups defined outside `groups.go` — `github`, `role`, `tag`, `webhook`, `attachment`, `mcp`, `playbook` — still exited 0 on typos. I extended your guard to cover those too, so all 16 groups are now consistent, and added a regression test that walks the *real* command tree and asserts every pure container group rejects unknown subcommands (so a future group can't silently regress). I rebased onto current `main` to pick up the file split.

Appreciate the contribution — this is a genuinely nice quality-of-life fix for anyone scripting or driving pad from an agent. Hope to see more PRs from you! 🚀

PR #895, opened 2026-07-10
fix(cli): silence usage on runtime errors, echo failing input

### LINKED ISSUE #852: CLI: stop dumping usage on runtime errors; echo the failing input
### Problem
When a command fails at runtime (e.g. item not found), the CLI prints the entire Cobra usage block — noisy — and never echoes what you asked for.

### Fix
- Set `SilenceUsage = true` on the root command (flag-parse errors still show usage; runtime errors won't).
- Wrap not-found errors with the input, e.g. `item TASK-999999 not found in workspace docapp`.

### Pointers
- `cmd/pad/main.go` — root command setup + error paths.

Small, high-signal polish.

b4rk13:
## What does this PR do?

Fixes #852.

Two changes to the CLI error experience:

1. **Silence usage on runtime errors** — sets `SilenceUsage: true` on the root Cobra command in `newRootCmd()`. Previously, any `RunE` error (item not found, server unreachable, etc.) caused Cobra to dump the full usage block before the error message, making it hard to see what went wrong.

2. **Echo the failing input in error messages** — wraps not-found / lookup errors with the ref and workspace that were queried (e.g. `item TASK-999999 not found in workspace myws`), so the user can see exactly what they asked for even when the error is terse.

**Note on flag-parse errors:** With `SilenceUsage: true`, Cobra silences usage for both runtime errors and unknown-flag errors. A custom `FlagErrorFunc` would be required to restore usage-on-flag-parse-error behavior. This PR accepts both-silenced as the intended outcome — the test file documents this explicitly.

## How to test

1. Build and install: `make install`
2. **Runtime error (no usage):** run `pad item get TASK-nonexistent` — you should see only the error message naming the ref and workspace, with no usage block.
3. **Unknown flag (no usage):** run `pad --nonsense-flag` — usage is also suppressed (both-silenced behavior per note above).
4. Run `go test ./cmd/pad/...` — `TestSilenceUsageBehavior` and `TestWrapNotFound` cover the two changes.

## Checklist

- [x] `make build` passes
- [x] `make test` passes
- [ ] New features have tests (if applicable) — yes: `TestSilenceUsageBehavior` + `TestWrapNotFound` added
- [ ] TypeScript types updated (if API changed) — N/A (CLI-only change)
- [ ] CLI help text updated (if new command) — N/A (no new commands)

### DIFF
--- a/cmd/pad/cmd_item.go
+++ b/cmd/pad/cmd_item.go
@@ -19,6 +19,19 @@ import (
 	"github.com/PerpetualSoftware/pad/internal/models"
 )
 
+// wrapNotFound converts a not-found APIError into a message that includes the
+// failing ref and workspace, e.g. "item TASK-999999 not found in workspace docapp".
+// All other errors pass through unchanged.
+func wrapNotFound(err error, ref, ws string) error {
+	if err == nil {
+		return nil
+	}
+	if apiErr, ok := err.(*cli.APIError); ok && apiErr.Code == "not_found" {
+		return fmt.Errorf("item %s not found in workspace %s", ref, ws)
+	}
+	return err
+}
+
 // parseFieldFlag parses a --field key=value flag value according to the
 // field's declared schema type. JSON-typed and multi_select fields receive
 // parsed JSON values, number-typed fields receive numbers, and checkbox
@@ -502,7 +515,7 @@ func showCmd() *cobra.Command {
 
 			item, err := client.GetItem(ws, args[0])
 			if err != nil {
-				return err
+				return wrapNotFound(err, args[0], ws)
 			}
 
 			// PLAN-1593 / TASK-1596: fetch the top 5 backlinks
@@ -897,7 +910,7 @@ Examples:
 						return fmt.Errorf("update rejected: item was modified by another writer")
 					}
 				}
-				return err
+				return wrapNotFound(err, slug, ws)
 			}
 
 			if formatFlag == "json" {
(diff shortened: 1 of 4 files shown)


xarmian:
Thank you for this, @b4rk13. Genuinely well-structured work, and the test that *documents* Cobra's real flag-parse behavior (rather than the issue's assumed split) is exactly the right instinct.

This one got accidentally swept up too: issue #852 was independently addressed in #893 (`c62658a1`), which merged while this was in draft. That's why GitHub is now showing conflicts. Sorry for the overlap, it landed before I connected it back to your PR. #893 took a slightly different tack on the two changes:
- **Usage:** `SilenceUsage: true` plus a `FlagErrorFunc`, so flag-parse errors still print usage (runtime errors don't), which is the split the original issue was reaching for.
- **Not-found wrap:** done at the client layer (`internal/cli/client.go`) returning an *enriched* `*cli.APIError`, to preserve the concrete error type. `bulk-update` asserts `err.(*cli.APIError)` to capture the code, so keeping the type intact mattered there.

So I'm going to close this PR, but I don't want your effort to evaporate, because you found something #893 *didn't* cover: the `pad item comment` path. #893 wraps `GetItem`/`UpdateItem`/`DeleteItem` but not `CreateComment`, so `pad item comment TASK-999999 "..."` still surfaces the bare not-found message. Would you be up for a small PR extending `wrapItemNotFound` (in `client.go`) to `CreateComment`? Same pattern as the others, and it closes the last gap. Happy to review it fast.

PR #900, opened 2026-07-10
refactor(web): migrate KeyboardShortcuts modal to the shared Modal primitive

### LINKED ISSUE #899: Web: migrate KeyboardShortcuts modal to the shared Modal primitive
Follow-up to #855. The shared native-`<dialog>` `Modal` primitive landed and the backdrop modals were migrated (TASK-2083, #890, #891). One genuine hand-rolled modal remains: `web/src/lib/components/common/KeyboardShortcuts.svelte` still uses a hand-rolled `<div class="backdrop" role="dialog" aria-modal="true">` instead of the shared primitive, so it misses focus trap/restore and top-layer stacking.

### Fix
Migrate `KeyboardShortcuts.svelte` to the shared `Modal` primitive, matching the pattern used by the already-migrated modals.

### Note
The other remaining `fixed inset-0` / `backdrop` overlays (Sidebar drawer, WorkspaceSwitcher popover, graph drawer, NotificationPanel, DockedSheet, Lightbox) are non-modal patterns and are intentionally out of scope here.

dchaudhari7177:
Closes #899

## What

`KeyboardShortcuts.svelte` was the last hand-rolled modal after #890/#891: a `<div class="backdrop" role="dialog" aria-modal="true">` with its own `svelte:window` Escape listener and backdrop-click handler — no focus trap/restore, no top-layer.

Migrated to the shared native-`<dialog>` `Modal` primitive, matching the already-migrated modals:

- `Modal` now owns Escape, backdrop dismiss, the focus trap, and focus save/restore; the component's own keydown/backdrop handlers are deleted. `visible` stays the single source of truth in `+layout.svelte` (its redundant Escape branch is harmless — both paths just flip the same state).
- The original look is preserved the same way `UserModal` did it: `placement="center"` (the old backdrop was flex-centered, unlike the primitive's `top` default) and `--modal-shadow="0 8px 32px rgba(0,0,0,0.3)"`; `bg/border/radius` already match the primitive's defaults. `maxWidth="500px"` carries over.
- The `.body` becomes the scroll area (`overflow-y: auto`) since the Modal box is `overflow: hidden` flex-column — previously the whole box scrolled.
- `aria-labelledby` is wired to the now-`id`'d `Keyboard Shortcuts` heading (the old version only had `aria-label` on the backdrop).

## Tests

New `KeyboardShortcuts.svelte.test.ts` in the jsdom project, following `Modal.svelte.test.ts`: renders through `dialog.modal` with the aria link and all three shortcut groups when `visible`; stays closed with content torn down when not; and the Close button calls `onclose` while leaving the dialog open until the parent flips `visible` (single-source-of-truth semantics).

## Verification

- `npm run check` — 0 errors (the 8 CSS warnings are pre-existing, all in unrelated route files)
- `npm test` — 190 passed (187 existing + 3 new)
- `grep` confirms no e2e spec targets the old `.backdrop` markup

### DIFF
--- a/web/src/lib/components/common/KeyboardShortcuts.svelte
+++ b/web/src/lib/components/common/KeyboardShortcuts.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+	import Modal from './Modal.svelte';
+
 	interface Props {
 		visible: boolean;
 		onclose: () => void;
@@ -20,17 +22,17 @@
 		{
 			title: 'Global',
 			shortcuts: [
-				{ key: '\u2318K', description: 'Search / Command palette' },
-				{ key: '\u2318N', description: 'New item' },
-				{ key: '\u2318\\', description: 'Toggle sidebar' },
+				{ key: '⌘K', description: 'Search / Command palette' },
+				{ key: '⌘N', description: 'New item' },
+				{ key: '⌘\\', description: 'Toggle sidebar' },
 				{ key: '?', description: 'Show keyboard shortcuts' }
 			]
 		},
 		{
 			title: 'Navigation',
 			shortcuts: [
-				{ key: 'j / \u2193', description: 'Move down' },
-				{ key: 'k / \u2191', description: 'Move up' },
+				{ key: 'j / ↓', description: 'Move down' },
+				{ key: 'k / ↑', description: 'Move up' },
 				{ key: 'Enter', description: 'Open selected item' },
 				{ key: 'Esc', description: 'Go back / Close' }
 			]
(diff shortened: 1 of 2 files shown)


xarmian (approved):
Thank you, @dchaudhari7177 — this is an excellent first contribution, and a fast one. 🙌

Reviewed and verified locally:
- Correct use of the shared `Modal` primitive, matching the `UserModal` reference migration (`open={visible}`, `{onclose}`, `labelledby`, `placement="center"`, `maxWidth`), and it correctly avoids wrapping `<Modal>` in `{#if}` per the primitive's contract.
- The `--modal-shadow` override is real (Modal reads `var(--modal-shadow, …)`), and moving the scroll to `.body` is the right call given the Modal box is `overflow: hidden`.
- Hand-rolled Escape/backdrop listeners correctly removed (the primitive owns them now), and `aria-labelledby` is wired to the heading — a genuine a11y improvement over the old `aria-label`-on-backdrop.
- `npm run check` clean (0 errors; the 8 warnings are pre-existing and unrelated), and your 3 new tests pass — nicely modeled on `Modal.svelte.test.ts`.

Approving. I'll squash-merge once CI goes green. Thanks again for picking this up so quickly!

PR #910, opened 2026-07-11
fix(cli): wrap comment not-found errors with item context

### LINKED ISSUE #902: CLI: wrap `pad item comment` not-found error with the failing ref + workspace
### Problem
#893 made item-not-found errors friendly by wrapping them with the failing ref and workspace (e.g. `item TASK-999999 not found in workspace docapp`). But the wrapper is only applied to `GetItem`, `UpdateItem`, and `DeleteItem`. `pad item comment TASK-999999 "..."` still surfaces the bare API not-found message, so the comment path is the odd one out.

### Fix
Extend the existing `wrapItemNotFound` helper to the `CreateComment` client path so `pad item comment` gets the same ref+workspace message as the other item commands. Same one-line pattern already used at `internal/cli/client.go:241/249/255`.

### Pointers
- `internal/cli/client.go` — `wrapItemNotFound` (line ~266) and `CreateComment` (line ~384).
- `internal/cli/item_notfound_test.go` — `wrapItemNotFound` is already table-tested; add a comment-path case.

cnYui:
## What does this PR do?

Applies the existing `wrapItemNotFound` behavior to `Client.CreateComment`, so `pad item comment TASK-999999 ...` reports the failing item ref and workspace instead of the bare API message.

The wrapper preserves the original `APIError` code, details, and concrete type, matching the existing Get/Update/Delete item paths.

Closes #902

## How to test

1. `go test ./internal/cli -run 'Test(GetItemWrapsNotFound|NotFoundPreservesDetails|GetItemPassesThroughOtherErrors)$' -count=1`
2. `go vet ./internal/cli`
3. `gofmt -d internal/cli/client.go internal/cli/item_notfound_test.go`
4. `git diff --check`

The focused tests pass on Windows. Broader local package runs have existing workspace prerequisites/platform failures (missing generated `web/build` embed output and Windows HOME/credentials assumptions), unrelated to this two-file CLI change.

## Checklist

- [ ] `make build` passes - not run locally; requires the generated web embed assets
- [ ] `make test` passes - broader Windows baseline has unrelated HOME/credentials failures
- [x] TypeScript types updated (if API changed) - N/A
- [x] CLI help text updated (if new command) - N/A

### DIFF
--- a/internal/cli/client.go
+++ b/internal/cli/client.go
@@ -383,7 +383,8 @@ func (c *Client) ListComments(wsSlug, itemSlug string) ([]models.Comment, error)
 
 func (c *Client) CreateComment(wsSlug, itemSlug string, input models.CommentCreate) (*models.Comment, error) {
 	var result models.Comment
-	return &result, c.post("/workspaces/"+wsSlug+"/items/"+itemSlug+"/comments", input, &result)
+	err := c.post("/workspaces/"+wsSlug+"/items/"+itemSlug+"/comments", input, &result)
+	return &result, wrapItemNotFound(err, itemSlug, wsSlug)
 }
 
 func (c *Client) DeleteComment(wsSlug, commentID string) error {
--- a/internal/cli/item_notfound_test.go
+++ b/internal/cli/item_notfound_test.go
@@ -65,6 +65,9 @@ func TestGetItemWrapsNotFound(t *testing.T) {
 	if err := client.DeleteItem("docapp", "TASK-999999"); err == nil || err.Error() != want {
 		t.Errorf("DeleteItem: got %v, want %q", err, want)
 	}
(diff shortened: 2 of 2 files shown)


xarmian (approved):
Reviewed and verified against the codebase — this is a clean, correct fix for #902.

- The one-line change matches the exact `wrapItemNotFound` pattern already used at `internal/cli/client.go:241/249/255`.
- `wrapItemNotFound` only rewrites `not_found` APIErrors and passes all other errors through unchanged, so wrapping the full `CreateComment` error introduces no behavior change for non-404 failures.
- Adds a real case to the existing table test.

The local Windows `make build`/`make test` gaps you noted are the expected `web/build` embed prerequisite, unrelated to this change. CI has been approved to run.

Thanks for the tidy, well-scoped contribution!

PR #911, opened 2026-07-11
ci: add Windows and macOS smoke tests

### LINKED ISSUE #857: CI: add a Windows + macOS smoke matrix
### Problem
Pad claims Windows and macOS support, but CI and release only run on `ubuntu-latest` — those platforms are never actually exercised.

### Fix
Add a small CI matrix on `windows-latest` + `macos-latest` that builds the binary and runs a smoke check: `pad --help`, boot the server, and one CLI round-trip (create + read an item). Optionally a post-release step that boots the built artifact.

### Pointers
- `.github/workflows/` — the existing CI workflow.

Self-contained; good for someone who likes CI.

raj921:
Fixes #857

Adds a native CI matrix for macOS and Windows. Each runner builds Pad, checks the help command, starts an isolated local server, and verifies an item create/read round-trip through the CLI.

Checks run:
- actionlint .github/workflows/ci.yml
- local macOS build and CLI smoke test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,6 +127,81 @@ jobs:
       - name: Verify binary runs
         run: ./pad --help
 
+  native-smoke:
+    name: Smoke (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    defaults:
+      run:
+        shell: pwsh
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-latest
+            binary: pad
+            path: ./pad
+          - os: windows-latest
+            binary: pad.exe
+            path: .\pad.exe
+    steps:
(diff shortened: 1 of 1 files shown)


xarmian (approved):
Reviewed and verified against the codebase — resolves #857 as specified.

Confirmed every command/env-var the smoke job uses actually exists:
- `/api/v1/health` endpoint (`server.go`), `PAD_BYPASS_SETUP_TOKEN` open-bootstrap path
- `pad server start --port`, `pad auth configure --mode local --port`
- `pad auth setup --email/--name/--password` (headless bootstrap), `pad workspace create <name> --slug --template`
- global `--workspace`/`--format`, `item create task` / `item show`, and item JSON carries `ref`

Nice security hygiene too — actions are pinned to full commit SHAs rather than floating tags, isolated temp data dir, no secrets touched. CI has been approved so the new Windows/macOS matrix can prove itself; will merge once it's green.

Thanks for picking this up!

PR #919, opened 2026-07-11
feat(cli): add pad item open command

### LINKED ISSUE #904: CLI: add `pad item open <ref>` to open an item in the browser
### Problem
`pad server open` opens the web UI, but there's no way to jump straight to a specific item from the CLI. When you're looking at `TASK-5` in the terminal, opening its web page means navigating by hand.

### Fix
Add `pad item open <ref>` that resolves the item and opens its web URL in the default browser. Reuse the browser-open helper behind `pad server open` (`cmd/pad/cmd_server.go::openCmd`) and the server base URL from `pad server info`. Should accept issue IDs (`TASK-5`) like every other item command.

### Pointers
- `cmd/pad/cmd_server.go::openCmd` — existing browser-open logic to reuse.
- `cmd/pad/cmd_item.go` — add the subcommand alongside the other item commands; wire it into the item group in `cmd/pad/groups.go`.

Rionlyu:
## What does this PR do?

Adds `pad item open <ref>` so a CLI user can jump directly from an item reference to its page in the Pad web UI.

- Resolves the supplied reference through the existing item client and fetches the workspace metadata needed for the web route.
- Builds the canonical item URL as `/{owner}/{workspace}/{collection}/{ref-or-slug}` from the configured `BrowserURL()`, using the canonical item ref when available and falling back to the item slug.
- Opens the item URL directly using the browser-opening helper behind `pad server open`.
- Returns item/workspace resolution and browser-launch errors without opening an incorrect page.

Closes #904

## How to test

1. Run `make check`.
2. Run `./pad item open --help` and confirm the command appears under `./pad item --help`.
3. With a configured workspace, run `./pad --workspace <workspace> item open TASK-5` and confirm the browser opens the canonical item page directly.

## Checklist

- [x] `make build` passes
- [x] `make test` passes
- [x] TypeScript types updated (if API changed) — no API change
- [x] CLI help text updated (if new command)

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,7 @@ pad project reconcile                 Reconcile item and PR state
 pad item create <coll> "title"        Create item (task, idea, plan, doc, ...)
 pad item list [collection]            List items (filters: --status, --priority, --all)
 pad item show <ref>                   Show item detail
+pad item open <ref>                   Open item in web UI
 pad item update <ref>                 Update item fields
 pad item delete <ref>                 Delete item
 pad item move <ref> <collection>      Move item between collections
--- a/cmd/pad/cmd_item.go
+++ b/cmd/pad/cmd_item.go
@@ -704,6 +704,42 @@ func showCmd() *cobra.Command {
 	}
 }
 
+// --- open ---
+
+func itemOpenCmd() *cobra.Command {
+	return itemOpenCmdWithOpener(openBrowser)
+}
+
+func itemOpenCmdWithOpener(opener func(string) error) *cobra.Command {
+	return &cobra.Command{
+		Use:   "open <ref>",
+		Short: "Open an item in the Pad web UI",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, cfg := getClient()
+			ws := getWorkspace()
+
+			item, err := client.GetItem(ws, args[0])
+			if err != nil {
+				return err
+			}
+
+			ref := cli.ItemRef(*item)
+			if ref == "" {
+				return fmt.Errorf("item %q has no canonical reference", args[0])
+			}
+
+			target := fmt.Sprintf("%s/-/r/%s/%s",
+				cfg.BrowserURL(),
+				url.PathEscape(ws),
+				url.PathEscape(ref),
+			)
+			fmt.Printf("Opening %s\n", target)
+			return opener(target)
+		},
+	}
+}
+
 // --- update ---
 
 func updateCmd() *cobra.Command {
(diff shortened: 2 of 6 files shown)


Rionlyu:
FYI, kept out of scope for this PR: the shared include-deleted item resolver does not preserve an old prefix after an item moves collections. For example, after PLAN-42 becomes TASK-42, GetItemByRef can still resolve PLAN-42 by workspace item number, while ResolveItemIncludeDeleted may not. This behavior predates pad item open and affects shared item reads, so no store change is included here; PR #919 remains focused on issue #904.

xarmian (commented):
Thanks for this, @Rionlyu. It's a clean, well-tested contribution and closes #904 nicely. The `itemWebPath()` helper and the test coverage are solid. A few things I'd like to see before merging:

**1. Open the item URL directly, and drop the `/login?redirect=` wrapper.**
Right now the command opens `{base}/login?redirect={item-path}` rather than the item URL itself. The web app already handles logged-out visitors: when an item page loads without a session, the API client bounces to `/login?redirect=<current-path>` and forwards back after auth (`web/src/lib/api/client.ts:467`). So the manual login wrapping duplicates that flow, and it costs the common case a hop, since a logged-in user (e.g. right after `pad server open`) gets routed through `/login` instead of landing on the item directly. Opening `baseURL + itemPath` gives you the same graceful logged-out handling for free, and lands authenticated users straight on the page. (For what it's worth, the server-side `/-/r/{ws}/{ref}` resolver isn't a good fit here either, since it returns a raw JSON 404 to logged-out browsers rather than redirecting to login.)

**2. Fix the print/open mismatch.**
```go
fmt.Printf("Opening %s%s\n", baseURL, itemPath)   // prints the item URL
...
return opener(loginURL)                            // opens /login?redirect=...
```
The command tells the user one URL and opens another. Dropping the login wrapper (point 1) resolves this automatically, since you'd print and open the same thing.

**3. Reconcile the PR description with the code.**
The description says it "builds a URL from the existing `/-/r/{workspace}/{ref}` resolver route," but the code constructs `/{owner}/{ws}/{coll}/{ref}` and wraps it in `/login?`. Looks like the description is from an earlier approach (the "make item open use canonical web routes" then "revert" churn). Mind updating it to match what the code does after the change above?

With those addressed this should be good to go. Thanks again!

Rionlyu:
Thanks for the clear review. 

The command now opens baseURL + itemPath directly. 
The printed and opened URLs use the same itemURL, with regression coverage.
The PR description now reflects the canonical web route.

Ready for another look when you have a chance.

xarmian (approved):
Looks great, thanks for the quick turnaround @Rionlyu. Opens the item URL directly now, print/open match, and the description lines up with the code. Builds clean and all tests pass locally. Merging.

PR #977, opened 2026-07-19
feat(web): retain-alive master freeze — peeking prop + complete mutation-path audit (TASK-2172)

xarmian:
## What

PLAN-2154 Phase 2 (Architecture E, decision **D2** + risks **R6/R9/R11/R12**). Adds a `peeking` prop to `ItemDetail` so the full-page master goes **read-only + ALIVE** while a detail pane peeks beside it — **retain-alive**: the collab provider is never torn down, and there is **no** flush / snapshot / persistence barrier / server ack. Only the mutation surfaces freeze.

`peeking` defaults **false**, so every existing caller (the collection page's pane `ItemDetail`, the full-page route wrapper) is **byte-identical**. No host passes `peeking={true}` yet — that wiring lands in **TASK-2174 (E)**.

## The freeze

`mutationsEnabled = canEdit && !peeking` (factored into the unit-tested `computeMutationsEnabled` helper) gates every local/user-originated mutation:

- title edit, fields, tags, assign user/role, move, delete
- relationship add/remove; the open-children **409 force-retry** callbacks for field + move recheck the freeze
- **ChildItems** add-child + link-child + reorder (child-row navigation stays live); an already-open add-child form unmounts on freeze
- the **raw-markdown editor** (`readonly={!canEdit || peeking}`) + its pending REST `item.content` save queue drained on peeking-begin (**R6**)
- the **Rich⇄Markdown mode toggle** — a provider-*lifecycle* control (Markdown nulls `collabKey` and destroys the provider); hidden + guarded, incl. a post-await recheck so a peek that begins mid-flush can't tear the provider down
- the rich **Editor**: `editable={!peeking}` **and** `peeking` in the `{#key}` so it remounts re-bound to the **same live Y.Doc**, shedding `BlockDragHandle` (closes the drag-reorder hole); slash menu / mobile+table toolbars auto-disable
- `EditorBubbleMenu` / `EditorLinkPopover`
- **AttachmentUpload** paste/drop DOM handlers + `uploadAttachments` command (ProseMirror runs DOM handlers even when `editable=false`) and the async completion path (rejects a destroyed/read-only view from the peek remount)
- **AttachmentImage** node toolbar (**R12**): rotate/crop + toolbar visibility gate on `editor.isEditable`, and `swapNodeUuid` re-checks editable/destroyed before its post-await Yjs dispatch
- **timeline** (via a new `frozen` prop → `ItemTimeline` → `TimelineCommentCard` / `TimelineVersionCard`): comment compose, reply/reaction/delete, already-open author/admin comment+reply edit forms (unmounted, taking their `CommentEditor` direct-upload), and version restore
- **archived restore** (`canRestore && !peeking`, since `canEdit` is forced false for archived items)

**R9** — the pane `onDestroy` nulls the shared `collectionStore.activeItem` only **clear-if-owner** (`activeItem?.id === item?.id`); the master re-asserts `setActiveItem(item)` on un-peek. **R11** — the master's tab-title effect yields while peeking (pane wins).

The collab-snapshot **background flush** is intentionally **not** gated — it fires only for remote sync into the still-live doc, which retain-alive expects (the acceptance carves out remote sync).

## Folded-in defect

`ChildItems` add/link-child ignored its own `canEdit` prop (only reorder honored it). `showAddChild` + `submitCreate`/`confirmLink` now gate on it, which also tightens the archived-item case.

## Tests

- `computeMutationsEnabled` truth table (`mutationGate.test.ts`)
- `FreezeProbe` mount test (`masterFreeze.svelte.test.ts`) asserting the gate wiring: with `peeking={true}` every mutation surface is unmounted / read-only and byte-identical to the `canEdit`-only baseline when not peeking

The **running-app mutation-silence e2e** (no local Yjs transaction / REST mutation fires against a real, peeking `ItemDetail`; remote collab sync stays fine) is deferred to **TASK-2175 (F)** — no host passes `peeking={true}` until TASK-2174, so it can't be exercised end-to-end yet.

## Collab lockstep

UI/behavioral freeze only — **not** a Y.Doc node-spec change. `SCHEMA_VERSION` / `DefaultSchemaVersion` unchanged; the persisted document shape is untouched (the peek remount re-binds the same Y.Doc).

## Gates

- `npm run check`: 0 errors
- `npx vitest run`: 442 passed (32 files)
- `npx playwright test pane- --workers=2`: 49 passed, 49 skipped (mobile-chromium project), 0 failures
- Codex review (xhigh): CLEAN after 3 rounds (R1 found 5, R2 found 3, R3 clean)

https://claude.ai/code/session_01EZ6yr6pAUFb1uffan912ra

### DIFF
--- a/web/src/lib/components/ChildItems.svelte
+++ b/web/src/lib/components/ChildItems.svelte
@@ -38,6 +38,18 @@
 		 */
 		canEdit?: boolean;
 		/**
+		 * PLAN-2154 Phase 2 / D2 / R12 (TASK-2172): master-freeze. When the
+		 * full-page host peeks a pane beside this item's ItemDetail, the master
+		 * passes `frozen={true}` so ALL ChildItems mutations (add/link-child +
+		 * reorder) freeze while child-row NAVIGATION stays live. Kept SEPARATE
+		 * from `canEdit` on purpose: add/link-child authorize off independent
+		 * target-collection / source-child capabilities (NOT the parent's edit
+		 * permission), so folding the freeze into `canEdit` would wrongly strip
+		 * add-child from a non-peeking user who lacks parent-edit but has those
+		 * capabilities. Defaults false → byte-identical for existing callers.
+		 */
+		frozen?: boolean;
+		/**
 		 * Instance-scoped shadow of the OWNING ItemDetail's own dirty/
 		 * lastSaveTime (PLAN-2154 Phase 0 / R4, TASK-2156). ChildItems used to
 		 * read the global `editorStore.dirty`/`lastSaveTime` singleton directly
@@ -60,7 +72,7 @@
 		onOpenTarget?: (target: PaneTarget) => void;
 	}
 
-	let { wsSlug, username = '', itemSlug, itemId, parentFields, terminalStatuses, onChildrenChange, canEdit = true, selfDirty = false, selfLastSaveTime = 0, onOpenTarget }: Props = $props();
+	let { wsSlug, username = '', itemSlug, itemId, parentFields, terminalStatuses, onChildrenChange, canEdit = true, frozen = false, selfDirty = false, selfLastSaveTime = 0, onOpenTarget }: Props = $props();
 
 	const defaultTerminal = ['done', 'completed', 'resolved', 'cancelled', 'rejected', 'wontfix', 'fixed', 'implemented', 'archived', 'disabled', 'deprecated'];
 	const terminal = $derived(terminalStatuses ?? defaultTerminal);
@@ -142,12 +154,23 @@
 		groupData[status] = e.detail.items;
 		isDragging = false;
 
+		// HT-2176 Option A (TASK-2172): gate reorder INITIATION, not finalization.
+		// A NEW drag can't START while frozen (`dragDisabled: !canEdit || frozen`
+		// on the dnd zone), but a drag already IN PROGRESS when the pane opens must
+		// finalize and persist — so do NOT reject on `frozen` here. `!canEdit`
+		// stays (the zone gate mirror; a non-editor never reaches finalize anyway).
+		if (!canEdit) return;
+
 		const updates = groupData[status]
 			.filter((i: any) => !i[SHADOW_ITEM_MARKER_PROPERTY_NAME])
 			.map((item, index) => ({ id: item.id, sort_order: index }));
 
 		try {
 			for (const { id, sort_order } of updates) {
+				// HT-2176 Option A (TASK-2172): NO per-PATCH freeze recheck. The
+				// reorder was INITIATED before peeking (the top guard blocks a NEW
+				// one); breaking mid-loop would persist it only partially, leaving
+				// inconsistent sort_orders. Let the initiated reorder finish.
 				await api.items.update(wsSlug, id, { sort_order });
 			}
 		} catch (e) {
@@ -162,6 +185,9 @@
 	// persists the changed rows via the same per-child update loop the drag
 	// path uses. A canonical reload (SSE/sync) settles it afterward.
 	async function reorderChild(status: string, child: Item, dir: ReorderDirection) {
+		// Freeze guard (TASK-2172 / R14): mirror handleFinalize. The kebab is
+		// hidden while `!canEdit || frozen`; this drops a straggler invocation.
+		if (!canEdit || frozen) return;
 		const grp = (groupData[status] ?? []).filter(
 			(i: any) => !i[SHADOW_ITEM_MARKER_PROPERTY_NAME]
 		);
@@ -174,6 +200,8 @@
 		const updates = reorderGroup(grp, child.id, dir);
 		try {
 			for (const u of updates) {
+				// Option A (TASK-2172): no per-PATCH freeze recheck — a reorder
+				// initiated pre-pane finishes fully (see handleFinalize).
 				await api.items.update(wsSlug, u.item.id, { sort_order: u.sort_order });
 			}
 		} catch (e) {
@@ -331,8 +359,11 @@
 		);
 	});
 
-	// Entry "+ Add child" shows iff at least one mode is usable.
-	let showAddChild = $derived(createTabEnabled || canLinkExisting);
+	// Entry "+ Add child" shows iff at least one mode is usable — gated on the
+	// INDEPENDENT target-collection / source-child capabilities (NOT the parent's
+	// `canEdit`), so the freeze must not touch that logic. `&& !frozen` layers the
+	// master-freeze on top without changing any non-peeking behavior (TASK-2172).
+	let showAddChild = $derived((createTabEnabled || canLinkExisting) && !frozen);
 
 	// ── Form state ─────────────────────────────────────────────────────────
 	let addOpen = $state(false);
@@ -468,6 +499,9 @@
 	}
 
 	async function submitCreate() {
+		// Freeze guard (TASK-2172): the form is hidden while frozen, but drop a
+		// straggler (e.g. an Enter keydown mid-freeze) so no child is created.
+		if (frozen) return;
 		const title = createTitle.trim();
 		const collSlug = createCollSlug;
 		if (!title || !collSlug || creating) return;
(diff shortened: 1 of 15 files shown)


xarmian:
Updated after the orchestrator's independent Codex pass, which found 8 issues the first loop missed (mostly R14 late-async-continuation freeze holes + one byte-identity regression). All fixed and re-verified — the freeze predicate is now rechecked IMMEDIATELY BEFORE every mutation dispatch (not just before the await):

1. **Provider retry** — `retryCollabSync` guarded on `!peeking` + its ContentError retry button hidden while peeking (`ContentError.onRetry` made optional). No provider teardown while peeking (retain-alive).
2. **Markdown-toggle flush loop** — in-loop fence now rechecks `peeking` before each flush dispatch.
3. **Raw REST queue** — peeking-begin no longer `flushNow()`s (a flush is itself a dispatch); it suppresses (cancel only) and re-flushes on un-peek. The raw `save` callback and `flushRawIfPending` drain recheck `mutationsEnabled` before every PATCH.
4. **Child reorder** — `handleFinalize` + `reorderChild` guard on `!canEdit || frozen` and recheck `frozen` before each per-row PATCH; drag/kebab gated on `frozen`.
5. **AttachmentImage crop** — rechecks `editor.isEditable`/`isDestroyed` after the modal resolves, before the server-side transform.
6. **Coalesced tag saves** — `flushTagSaver` rechecks `mutationsEnabled` before the drained dispatch.
7. **Import metadata** — `stampSourceUrl` rechecks `mutationsEnabled` after its GET, before the field PATCH.
8. **Byte-identity regression (important)** — ChildItems now takes the REAL `canEdit` + a SEPARATE `frozen={peeking}` prop. add/link-child authorize off INDEPENDENT target-collection / source-child capabilities, so `showAddChild` reverts to `(createTabEnabled || canLinkExisting) && !frozen`. A non-peeking user who can create/link but can't edit the parent keeps add-child. `mutationsEnabled === canEdit` when not peeking, everywhere.

Gates: `npm run check` 0 errors · `npx vitest run` 442 passed · `npx playwright test pane-` 49 passed / 49 skipped / 0 fail. Codex (xhigh) CLEAN verifying each of the 8.

xarmian:
Second orchestrator review round (3 findings, one my earlier fix introduced) — all fixed, plus a follow-up the Codex loop caught. Same discipline: recheck the freeze predicate immediately before the write; the freeze SUSPENDS then RESUMES a pending user mutation, never drops it.

1. **Coalesced tag edit was STRANDED (data loss) — introduced by the finding-#6 fix.** `flushTagSaver` now breaks on `!mutationsEnabled` leaving `pending` set, skips `showSaved()` (guarded on `pending===null`), and resets `saveStatus`; the peeking→false transition RE-FLUSHES every tag saver with a held `pending` (suspend-then-resume, mirroring the raw REST queue). No pending tag edit is lost.
2. **star / Share / owner quick-action config still dispatched REST while peeking.** Now trigger-gated in ItemDetail's chrome: star `disabled={peeking}` + onclick guard; Share `{#if isOwner && !peeking}`; `QuickActionsMenu canEdit={isOwner && !peeking}`; and the peeking-begin transition dismisses any already-open share / edit-collection / move / add-link surface. These gate on `peeking` (NOT `mutationsEnabled`) so a non-peeking VIEWER can still star and a non-peeking ARCHIVED-item owner keeps Share/quick-actions — byte-identical. Follow-up (Codex): `QuickActionsMenu`'s already-open "New quick action" create form is now gated `{#if showCreateForm && canEdit}` and `handleSaveNewAction` early-returns on `!canEdit`, closing the last `api.collections.update` path.
3. **R9 teardown byte-identity on a cross-collection load failure.** `onDestroy` clear-if-owner now compares `activeItem?.id` against `activeItemOwnedId` (the id this instance actually set active), not the reactive `item` (which a failed cross-collection load leaves ahead of `activeItem`) — so it clears the global `activeItem` exactly as the pre-R9 unconditional clear did.

Gates: `npm run check` 0 errors · `npx vitest run` 443 passed · `npx playwright test pane-` 49 passed / 49 skipped / 0 fail. Codex (xhigh) CLEAN.

xarmian:
Reworked to Dave's **HT-2176 Option A** and hardened across several independent review rounds. The freeze now blocks only the INITIATION of a NEW edit while peeking; a save/action the user started before the pane opened completes normally (no suppression, no data loss, no provider teardown).

**Removed** the pre-pane pending-save suppression machinery (tag suspend + un-peek re-flush; raw peeking-begin flushNow/cancel + raw `save`/`flushRawIfPending` rechecks; `stampSourceUrl`, force-retry, ChildItems reorder per-PATCH, and `handleVersionRestore`-adopt mid-flight rechecks). Confirmed the collab flush never tears down the provider (`createCollabFlusher.save` is only `api.items.flushCollabContent`; `provider.destroy` lives solely in the `{#key}`/forceRefreshNonce cleanup).

**Kept/hardened** the core NEW-action freeze: title/fields/tags/assign/move/delete, relationship-add, ChildItems add/link/reorder-initiation, timeline compose/edit/upload, version-restore-initiation, archived restore, star/Share/quick-actions triggers, the Rich⇄Markdown toggle (save completes, mode-flip blocked so the provider is never torn down), and `editable={!peeking}` + `peeking` in the Editor `{#key}`.

**R12 complete editor-DOM-handler audit** — ProseMirror runs `handleDOMEvents` before its editable check, so every custom handler that dispatches a mutation is now gated on `editable`: AttachmentUpload paste/drop/command, AttachmentImage rotate/crop/toolbar, the **code-block + table clipboard plugins' CUT (delete) and TSV PASTE (replace)** — copy stays read-only — and htmlBlock edit-mode entry. BlockDragHandle is shed by the remount; the slash/link keydown pickers can't focus a read-only editor.

Two documented, intentional tradeoffs: the D2 editor remount interrupts an in-flight editor-bound action (upload/transform/refresh) that degrades gracefully with no content loss; and the `editable`-gates also close the same latent holes for view-only viewers (whose attachment/cut/paste mutations were already broken — a bug fix). Byte-identical for all working non-peeking flows.

Gates: `npm run check` 0 errors · `npx vitest run` 443 passed · `npx playwright test pane- --workers=1` 49/49 (the one intermittent failure at `--workers=2` is the pre-existing `pane-content-link-anchors:230` graph-render timing flake — passes 3x in isolation; my changes are byte-identical when not peeking). Codex (xhigh) CLEAN.

xarmian:
Round-5 tidy-up (HT-2176 Option A boundary). Two fixes, two documented edges.

**Fixed:**
1. **Pre-pane debounced FIELD save now completes.** `updateField` no longer rechecks `mutationsEnabled` — a value the user typed before the pane opened (FieldEditor's 500ms debounce fires after peeking begins) still saves. NEW field input stays blocked at the UI (`FieldEditor readonly={!mutationsEnabled}`). Locked in with a `FieldSaveProbe` mount test that types, flips the field read-only mid-edit, and asserts the debounced `onchange` still fires with the typed value.
2. **A child reorder started before peeking now finalizes + persists fully.** `handleFinalize` no longer rejects on `frozen` — reorder is gated at INITIATION (`dragDisabled: !canEdit || frozen`), not finalization, so a drag already in progress when the pane opens persists all rows.

**Documented (no behavior change):**
3. The Rich⇄Markdown mode FLIP is deliberately refused while peeking — it would teardown/recreate the collab provider (retain-alive/D2). The content still flushes; only the flip defers. Comments added at both toggle handlers.
4. **BUG-2177** (accepted, tracked, sev=low): the D2 `{#key}` editor remount orphans an in-flight editor-bound action (upload/rotate-crop/source-refresh/timeline-composer-upload) — bails gracefully via `view.isDestroyed`/identity (no crash, no committed-content loss). BUG-2177 referenced at every bail site.

Gates: `npm run check` 0 errors · `npx vitest run` 444 passed · `npx playwright test pane- --workers=1` 49/49 (workers=1 per the known TASK-2159 graph-render parallel-load flake). Codex (xhigh) CLEAN — verified both fixes, both documented edges, byte-identity when not peeking, and that `view.isDestroyed`/`view.editable` are real ProseMirror properties backing the guards.

PR #1037, opened 2026-07-27
feat(nix): add flake packaging with CI build

jairbj:
## What does this PR do?

Adds Nix flake packaging for `pad`, with an automated CI build.

- `flake.nix` — exposes `packages.default`/`pad` (the binary), `apps.default`
  (`nix run`), a `devShells.default` (Go + Node pinned to the same versions
  CI uses), and `checks` (package build + `pad --version` smoke test).
- `nix/package.nix` — `buildGoModule` + `importNpmLock` derivation, written
  nixpkgs-submission-ready (no flake-specific inputs). Composes the
  SvelteKit build and the Go build via `postPatch` so `//go:embed
  all:web/build` in `embed.go` has real files to embed.
- `.github/workflows/nix.yml` — runs `nix flake check` + `nix build` on
  every push/PR, following the repo's existing SHA-pinned Actions
  convention.
- `README.md` — documents `nix run` / `nix profile install` / `nix develop`.
- Two small test fixes (`internal/server/handlers_webhooks_secret_test.go`,
  `internal/mcp/dispatch_http_routes_extras_test.go`): swapped a couple of
  `example.com` webhook URLs for a literal IP, matching the pattern already
  used in `handlers_webhook_token_idor_test.go`. `ValidateWebhookURL` does a
  real DNS lookup as an SSRF guard, which fails under Nix's sandboxed
  (network-less) build — same fix, extended to the two spots that still hit
  it.

## How to test

1. `nix build github:PerpetualSoftware/pad` (once merged), or from this
   branch: `nix build .#default`
2. `./result/bin/pad --version` / `--help`
3. `nix flake check` — builds the package (which runs `go test ./...` in
   `checkPhase`) and a version smoke test

## Checklist

- [x] `make build` passes
- [x] `make test` passes
- [x] New features have tests (if applicable) — N/A, packaging only; the two test
  tweaks preserve existing coverage, just swap the URL literal
- [x] TypeScript types updated (if API changed) — N/A, no API changes
- [x] CLI help text updated (if new command) — N/A, no new commands

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,38 @@
+name: Nix
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: nix-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+# All third-party Actions are pinned to a 40-char commit SHA with a trailing
+# '# vX.Y.Z' comment so a compromised maintainer or moved tag cannot silently
+# execute attacker code in CI. Bump the SHA + comment together when updating.
+
+jobs:
+  nix:
+    name: Nix build & check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: DeterminateSystems/nix-installer-action@21a544727d0c62386e78b4befe52d19ad12692e3 # v22
+
+      - name: nix flake check
+        run: nix flake check --print-build-logs
+
+      - name: nix build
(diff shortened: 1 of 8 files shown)


xarmian:
Merged — thank you @jairbj, this is careful work: the fileset-based src, importNpmLock to dodge the npmDepsHash treadmill, the sandbox-aware test handling, and matching the repo's SHA-pinning convention were all appreciated on review. Two maintainer tweaks landed on your branch before merge: (1) the nix-installer-action pin was ~68 commits behind the v22 tag it referenced, now points at the actual v22 commit; (2) vendorHash refreshed — go.sum on main had moved since you computed it. `nix run github:PerpetualSoftware/pad` is now real. If you're still up for the nixpkgs submission mentioned in the derivation comments, we'd welcome it. 🙏

Issue #898, opened 2026-07-10
CLI: implement `markdown` output for the list commands


xarmian:
### Background

Follow-up to #851 (option (a)). #851 validated the `--format` enum and made the help text honest about where `markdown` is actually supported (shipped in #893). This issue covers the other half: actually implementing `markdown` output everywhere the flag advertises it, so the format becomes honestly global rather than honestly-partial.

Today `markdown` is implemented on a handful of commands (e.g. `item show`, `project changelog`) but not on the list commands. The flag help currently reads: `table, json (markdown on select commands, e.g. item show, project changelog)`.

### Fix

Implement `markdown` rendering for the list commands that lack it:
- `pad item list`
- `pad collection list`
- (and any other list-style command missing `markdown`)

Once `markdown` is supported on the list commands, update the `--format` flag help in `cmd/pad/main.go` to advertise `markdown` as a first-class global format again.

### Pointers

- `cmd/pad/main.go` — the persistent `--format` flag + output helpers.
- Look at how `item show` / `project changelog` implement their `markdown` branch for the pattern to follow.

/cc @b4rk13 — this is the option (a) follow-up you proposed on #851.

b4rk13:
Taking this one. Thanks for the write-up on #851; no harm done on the overlap.

It's built already, on `feat/markdown-list-output` off `main` @ `be7baa7`. PR to follow shortly. What's covered:

- `pad item list`, grouped as `## Icon Name (N)` with a table per collection when listing across collections, and a single table when scoped to one. Heading style borrowed from `project changelog --format markdown` so the two read the same.
- `pad item starred`
- `pad collection list`
- A new `internal/cli/format_markdown.go` holding the renderers, so the next command that wants markdown is a two-line wiring change.

Two deliberate choices, both cheap to reverse:

1. The markdown renderers don't reuse `ColorizedStatus` / `PriorityColor` / `Dim`. Markdown goes to a file, a PR body, or an agent's context and never to a terminal, so raw values go in and the reader's renderer styles them. The side effect is that the status icon (`◐`, `✓`) is dropped in markdown; happy to keep it if you'd rather.
2. Every cell goes through an escape helper. An unescaped `|` in an item title silently adds a column and corrupts the row, which is the sort of thing that surfaces in someone's PR description weeks later.

[...]

PR #1070, opened 2026-08-10
feat(cli): markdown output for the list commands

b4rk13:
## What does this PR do?

Implements #898: `--format markdown` on the list commands that lacked it.

- **`pad item list`** — grouped `## Icon Name (N)` sections with a markdown table each when listing across collections, mirroring the table format's grouping; a single table when scoped to one collection. Heading style matches `project changelog --format markdown`.
- **`pad item starred`** — single table.
- **`pad collection list`** — Name / Slug / Items / Default.
- **`--format` help** — drops the "markdown on select commands" caveat added in #893.

Renderers live in a new `internal/cli/format_markdown.go` (`RenderItemMarkdown`, `RenderCollectionMarkdown`, `escapeMarkdownCell`), so wiring the next command is two lines.

Two design notes:

1. **No ANSI, by design.** The markdown path deliberately does not reuse `ColorizedStatus` / `PriorityColor` / `Dim`. Markdown output is destined for a file, a PR body, or an agent's context, so raw field values go in and the reader's renderer does the styling. A test forces `color.NoColor = false` and asserts the output contains no escape sequences, so a future edit can't quietly leak colour into a markdown table.
2. **Every cell is escaped.** An unescaped `|` in an item title silently adds a column and corrupts the row; newlines end it early. `escapeMarkdownCell` strips SGR escapes, escapes pipes, and collapses newlines.

Item columns mirror the table renderer's (Ref, Title, Status, Priority, Collection, Updated) so switching format changes the styling, not the information. Where colour carried meaning and can't survive, a glyph stands in: pinned items get a `📌` prefix on the ref (the table uses a yellow `*`), and archived items keep the `(archived)` suffix.

## Empirical output

`pad item list tasks --format markdown`:

| Ref | Title | Status | Priority | Collection | Updated |
| --- | --- | --- | --- | --- | --- |
| 📌 TASK-314 | Push to fork, draft PR | done | high | ✓ Tasks | 2 hours ago |
| BUG-42 | parser chokes on a \| b input | ready | — | 🐛 Bugs | 1 day ago |
| IDEA-7 | retired idea (archived) | — | — | 💡 Ideas | 3 days ago |

`pad collection list --format markdown`:

| Name | Slug | Items | Default |
| --- | --- | --- | --- |
| ✓ Tasks | tasks | 208 | yes |
| 💡 Ideas | ideas | 159 | yes |
| Decisions | decisions | 4 |  |

Empty result sets print `No items found.` / `No collections found.`, matching the table renderers.

## How to test

1. `go test ./internal/cli/ -run Markdown -count=1` and `go test ./cmd/pad/ -run GroupedByCollectionMarkdown -count=1`
2. `pad item list --format markdown` (grouped), `pad item list tasks --format markdown` (single table), `pad collection list --format markdown`, `pad item starred --format markdown`
3. `pad --help | grep format` for the updated flag help
4. Corrupt-table check: create an item with a `|` in the title and confirm the row still renders as six cells

## Gate results

- `go build ./...` — PASS
- `go vet ./internal/cli ./cmd/pad` — PASS
- `golangci-lint run ./internal/cli/... ./cmd/pad/...` — PASS (0 issues)
- `gofmt -l` on the changed files — clean
- New tests — PASS (8 in `internal/cli/format_markdown_test.go`, 2 in `cmd/pad/format_markdown_list_test.go`), each written before the implementation and confirmed failing first

## Checklist

- [ ] `make build` — not run; needs the generated `web/build` embed assets. `go build ./...` passes.
- [ ] `make test` — the two touched packages have a red Windows baseline on clean `main` (7 credential/HOME failures in `internal/cli`, plus `TestWriteFileAtomic_WritesContent` and `TestHeadlessSetupNoTokenFileOK` in `cmd/pad`). I captured that baseline by stashing this branch and re-running; the failure set is identical with and without these changes. Same wall @cnYui hit on #910.
- [x] New features have tests
- [x] TypeScript types updated — N/A, CLI-only
- [x] CLI help text updated — yes, the `--format` flag help

There's an open question on #898 about whether the remaining table-printing surfaces (`item deps`, `item comments`, `project activity`, `attachment list`, `library list`, `role list`, `workspace members`) should come in here too, which would let the flag help drop the parenthetical entirely. Happy to extend this PR either way.

### DIFF
--- a/cmd/pad/cmd_collection.go
+++ b/cmd/pad/cmd_collection.go
@@ -55,6 +55,11 @@ func collectionsCmd() *cobra.Command {
 				return cli.PrintJSON(colls)
 			}
 
+			if formatFlag == "markdown" {
+				cli.PrintCollectionMarkdown(colls)
+				return nil
+			}
+
 			cli.PrintCollectionTable(colls)
 			return nil
 		},
--- a/cmd/pad/cmd_item.go
+++ b/cmd/pad/cmd_item.go
@@ -397,15 +397,23 @@ Examples:
 				return cli.PrintJSON(cli.ToItemSummaries(items))
 			}
 
-			if len(items) == 0 {
+			// Grouped when listing across collections, a single table when
+			// scoped to one — markdown mirrors the table layout.
+			grouped := len(args) == 0 && groupBy == ""
+
+			switch {
+			case formatFlag == "markdown":
+				if grouped {
+					renderItemsGroupedByCollectionMarkdown(os.Stdout, items)
+				} else {
+					cli.PrintItemMarkdown(items)
+				}
+			case len(items) == 0:
 				fmt.Println("No items found.")
 				return nil
-			}
-
-			// Group by collection if listing all
-			if len(args) == 0 && groupBy == "" {
+			case grouped:
 				printItemsGroupedByCollection(items)
-			} else {
+			default:
 				cli.PrintItemTable(items)
 			}
 
@@ -439,6 +447,45 @@ Examples:
 	cmd.Flags().StringArrayVarP(&fieldFlags, "field", "f", nil, "filter by field value (repeatable): --field key=value")
 
 	return cmd
+}
+
+// renderItemsGroupedByCollectionMarkdown is the markdown counterpart of
+// printItemsGroupedByCollection (#898): one `## Icon Name (N)` heading per
+// collection, each followed by its own markdown table. Grouping is kept rather
+// than flattened so the markdown form carries the same information as the
+// table. Heading style matches `project changelog --format markdown`.
+func renderItemsGroupedByCollectionMarkdown(w io.Writer, items []models.Item) {
+	if len(items) == 0 {
+		fmt.Fprintln(w, "No items found.")
+		return
+	}
+
+	groups := make(map[string][]models.Item)
+	order := []string{}
+	for _, item := range items {
+		key := item.CollectionSlug
+		if _, exists := groups[key]; !exists {
+			order = append(order, key)
+		}
+		groups[key] = append(groups[key], item)
+	}
+
+	for i, key := range order {
+		groupItems := groups[key]
+		icon := ""
+		name := key
+		if groupItems[0].CollectionIcon != "" {
+			icon = groupItems[0].CollectionIcon + " "
+		}
+		if groupItems[0].CollectionName != "" {
+			name = groupItems[0].CollectionName
+		}
+		if i > 0 {
+			fmt.Fprintln(w)
+		}
+		fmt.Fprintf(w, "## %s%s (%d)\n\n", icon, name, len(groupItems))
+		cli.RenderItemMarkdown(w, groupItems)
+	}
 }
 
 func printItemsGroupedByCollection(items []models.Item) {
@@ -3079,6 +3126,11 @@ func starredCmd() *cobra.Command {
 
 			if formatFlag == "json" {
 				return cli.PrintJSON(items)
+			}
+
+			if formatFlag == "markdown" {
+				cli.PrintItemMarkdown(items)
+				return nil
 			}
 
 			if len(items) == 0 {
--- a/cmd/pad/cmd_item.go
+++ b/cmd/pad/cmd_item.go
@@ -472,18 +472,22 @@ func renderItemsGroupedByCollectionMarkdown(w io.Writer, items []models.Item) {
 
 	for i, key := range order {
 		groupItems := groups[key]
-		icon := ""
-		name := key
-		if groupItems[0].CollectionIcon != "" {
-			icon = groupItems[0].CollectionIcon + " "
-		}
+		// Sanitized, not raw: a newline or SGR sequence in a collection name or
+		// icon would otherwise inject document structure into the heading, which
+		// is the same hole the per-cell escaping closes inside the table.
+		// Sanitize the parts BEFORE joining, since sanitizing trims and would
+		// otherwise eat the separating space.
+		label := cli.SanitizeMarkdownText(key)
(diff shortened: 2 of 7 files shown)


xarmian (changes requested):
Thanks — this is a high-quality contribution: the renderer split, the forced-color ANSI test, and the empirical output in the PR body all made review straightforward. Verified locally on Linux: full build green, all 10 new tests pass, and both touched packages pass their complete suites (consistent with the Windows-only baseline you documented — thanks for capturing it against clean `main`).

Three changes requested, all small:

1. **Escape backslashes before pipes in `escapeMarkdownCell`.** A title containing `\|` currently becomes `\\|` — GFM reads that as a literal backslash followed by an *unescaped* pipe, so the row still corrupts. Adding `strings.ReplaceAll(s, "\\", "\\\\")` ahead of the pipe replacement closes it; a regression test alongside `EscapesPipesInTitle` would lock it in.

2. **Sanitize the group headings.** `renderItemsGroupedByCollectionMarkdown` interpolates the collection icon/name into `## …` raw. A collection name containing a newline (or a stray SGR sequence) can inject document structure, which cuts against the PR's own every-cell-is-escaped guarantee. Running the heading text through the SGR-strip + newline-collapse steps (pipe-escaping isn't needed outside a table) is enough.

3. **Tighten the `--format` help wording.** "markdown on the list/show commands" reads as *all* list commands, but `role list --format markdown` (and the other surfaces you enumerated on #898) still silently print tables. Per the answer over on #898 (option (a)), keep this PR's scope and make the wording precise instead, e.g.: `output format: table, json, markdown (markdown on: item list/starred, collection list, item show, project changelog)`.

Nit, take or leave: empty `item starred --format markdown` prints "No items found." where the table path says "No starred items." — and a small end-to-end test through the cobra command (rather than the render helpers) would cover the format routing itself.

b4rk13:
All three done, plus both nits. Pushed as `81222a3`.

**1. Backslash before pipe.** Good catch, and it was worse than cosmetic; I wrote the failing test first to be sure I understood it. `a \| b` was coming out as `a \\| b`, so GFM rendered an escaped backslash and then a live pipe, and the row gained a column exactly as you said. Now escaped backslash-first (`a \\\| b`), which renders back to the literal title. Order matters and is worth the comment I left in the code: escaping pipes first would then double-escape the backslashes the pipe-escape had just introduced.

**2. Heading sanitization.** You were right that this cut against the PR's own guarantee. Extracted `SanitizeMarkdownText` (SGR strip plus newline collapse) and the heading now runs the icon and the collection name through it. Two details worth flagging:

- I sanitize each part *before* joining them, because sanitizing trims, and doing it after would eat the space between the icon and the name.
- Pipes are deliberately left alone there, per your note that they aren't special outside a table.

Two tests cover it: a collection name carrying `\n## Injected Heading` now produces one heading rather than two, and a name wrapped in an SGR sequence comes out clean. The ANSI case was a nice illustration of the gap; the table *cell* was already stripping it correctly while the heading directly above it wasn't.

**3. Help wording.** Taken verbatim: `output format: table, json, markdown (markdown on: item list/starred, collection list, item show, project changelog)`. Agreed on the reasoning; "list/show commands" implied a promise `role list` doesn't keep.

**Nits, both taken.**

Empty `item starred --format markdown` now says "No starred items." The empty check moved above the format branch so both paths give the same message.

On the end-to-end test: done, and I'm glad you pushed for it, because the format routing was the one thing genuinely untested. `format_markdown_routing_test.go` drives `item list` and `collection list` through cobra against an httptest server and asserts three things: the markdown branch is reached, the grouped headings and tables appear, and the two paths don't leak into each other. I proved it catches something by disabling the markdown branch in `listCmd` and watching it fail, rather than trusting a test that passed first time.

It follows the `item_open_test.go` pattern, with one deliberate difference: I set `USERPROFILE` alongside `HOME`. `os.UserHomeDir` reads `USERPROFILE` on Windows and the credential store resolves through it, so a test that sets only `HOME` still points at the developer's real `~/.pad` there. That may be worth a look independently of this PR: it's a plausible cause of some of the credential-store failures in the Windows baseline I documented, and if so it's a test-environment issue rather than a product one. Happy to file it separately.

Gates after the changes: `go build ./...`, `go vet`, `gofmt`, and `golangci-lint` (0 issues) all clean, all markdown tests pass, and both touched packages show the same 6+2 pre-existing Windows failures as clean `main` under an identical sandboxed run.

On scope: understood on (a), and the reasoning makes sense to me. I'll open the follow-up for the remaining seven surfaces (`item deps`, `item comments`, `project activity`, `attachment list`, `library list`, `role list`, `workspace members`) once this lands, which is also the natural place to extend the routing test to cover each one. At that point the flag help collapses to `table, json, markdown` and #898 closes.

xarmian:
Merged — thanks for the fast, test-first turnaround on the review round. The seven-surface follow-up we agreed on over in #898 is very welcome whenever you feel like it; the renderer split you built should make it mostly two-line wiring. (If you want to fold in a small extra: #1076 tracks broadening the shared `sgrPattern` helper beyond SGR sequences — it touches the same code you'd be in anyway.)

b4rk13:
Thanks for the merge, and for a review round that actually improved the thing — the backslash case was a real bug and I'd have shipped it.

Follow-up is up as #1080: the seven remaining surfaces, plus #1076 folded in as you suggested. It was indeed mostly two-line wiring once there was a shared table helper, so the interesting parts turned out to be the two surfaces that aren't tabular (`item comments` and `item deps`) and one judgement call on #1076 that I've flagged in the PR body: I extended the broader pattern to `displayWidth` as well, since a zero-width control sequence being counted as visible columns is the same bug wearing a different hat. Easy to split back out if you'd rather.

PR #1080, opened 2026-08-12
feat(cli): markdown output on the remaining list surfaces; broaden ANSI stripping

### LINKED ISSUE #898: CLI: implement `markdown` output for the list commands

### LINKED ISSUE #1076: CLI: ANSI stripping only covers SGR sequences, not all escape/control sequences
`internal/cli/format.go`'s `sgrPattern` matches only `ESC[...m` (SGR). Non-SGR CSI sequences (e.g. `ESC[2K`), OSC-8 hyperlinks, and other control sequences survive "ANSI stripping" — including in the markdown renderers added in #1070, whose doc comments say ANSI escapes are stripped.

Data-borne only (a stored title would have to carry the sequence), so low severity. Fix: broaden the pattern (or add a companion) to strip CSI + OSC + stray C0 controls, with tests using non-SGR inputs.

Candidate to fold into the #898 seven-surface follow-up since it touches the same renderer code. While in there, the other small gap from the same review round: end-to-end routing tests for `item starred --format markdown` and the scoped `item list <collection> --format markdown` path.

https://claude.ai/code/session_01VxyZv1g6W6rGx7nuaGcH3i

b4rk13:
## What does this PR do?

Completes #898 and fixes #1076.

**The seven surfaces left out of #1070** now support `--format markdown`, so the format is honestly global and the flag help collapses to `output format: table, json, markdown`:

`item comments`, `item deps`, `project activity`, `attachment list`, `library list`, `role list`, `workspace members`.

Two of those aren't tabular, and markdown follows the terminal shape rather than forcing a table onto them:

- **`item comments`** keeps the attribution-line-then-body form, and the body is emitted **verbatim**. A comment body is authored as markdown, so escaping it would turn its lists and code fences into literal text — the opposite of what the format is for. Only the attribution line, which we construct, is sanitized.
- **`item deps`** keeps its two sections as `## Blocks` / `## Blocked by` lists. Colour carried the direction in the terminal (yellow `->` out, red `<-` in); headings carry it here.

**New shared spine:** `cli.RenderMarkdownTable(w, headers, rows)`. Every cell is escaped, and a row shorter or longer than the header is padded or truncated to the header width, so a ragged row can't shift the column count and stop the whole block parsing as a table. Wiring a surface is now naming columns and mapping rows, which is what made this diff mostly mechanical.

**#1076 — ANSI stripping.** `sgrPattern` matched only `ESC[…m`, so non-SGR CSI sequences, OSC-8 hyperlinks, and stray C0 controls survived. Replaced with `ansiPattern` + `stripANSI` covering OSC (to BEL or ST), CSI, two-character Fe escapes, and stray C0/DEL. TAB, LF, and CR are deliberately preserved — they're meaningful whitespace, and the callers that need them normalized (`SanitizeMarkdownText`) handle them explicitly.

One judgement call worth flagging: **`displayWidth` now uses the broader pattern too.** The issue framed this around the markdown renderers, but a control sequence is zero-width, so counting it as visible columns was a table-alignment bug of the same family. It only bites on data-borne sequences, same as the markdown case. Happy to split that back out if you'd rather keep this PR to the markdown surface.

## Empirical output

`pad item deps TASK-5 --format markdown`:

```markdown
# Dependencies for TASK-5 Ship the follow-up

## Blocks

- TASK-9 Docs refresh

## Blocked by

_none_
```

`pad item comments TASK-5 --format markdown` (note the body's own markdown surviving intact):

```markdown
**Dave (dave@example.com)** · just now via cli

Confirmed on Linux:

- build green
- `role list | wc -l` unchanged
```

`pad role list --format markdown`:

| Slug | Name | Description | Tools | Items |
| --- | --- | --- | --- | --- |
| implementer | 🔧 Implementer | Writes code, builds features | edit,bash | 12 |

`pad project activity --format markdown`:

| When | Actor | Action | Item | Changes |
| --- | --- | --- | --- | --- |
| 2026-08-11 09:14 | Claude | updated | TASK-440 Markdown follow-up | status: ready → done |

`library list` gets a `# Conventions` / `# Playbooks` heading with a `## category` heading and table per category; `workspace members` gets the member table plus a `## Pending invitations` table when there are any.

## How to test

1. `go test ./internal/cli/ ./cmd/pad/ -run 'Markdown|StripANSI|DisplayWidth' -count=1`
2. `pad item comments <ref> --format markdown` — confirm a body containing a list or a fenced block renders as markdown, not escaped text
3. `pad item deps <ref> --format markdown`, `pad project activity --format markdown`, `pad role list --format markdown`, `pad workspace members --format markdown`, `pad attachment list --format markdown`, `pad library list --format markdown`
4. `pad --help | grep format` — now just `table, json, markdown`
5. Non-SGR check: store a title containing `ESC[2K` and confirm it neither appears in markdown output nor misaligns the table

## Gate results

- `go build ./...` — PASS
- `go vet ./cmd/pad ./internal/cli` — PASS
- `gofmt -l` — clean
- `golangci-lint run ./internal/cli/... ./cmd/pad/...` — PASS (0 issues)
- New tests — 12 ANSI-stripping cases, 4 table-helper cases (including ragged rows), 4 renderer cases for the two non-tabular surfaces, and the routing test extended to **8 subtests, one per surface**, each driven through cobra against an httptest server. That covers the two gaps you named in #1076: `item starred` and the scoped `item list <collection>` path.
- Every new guard was proven by mutating the source and watching it fail, rather than trusting a test that passed first time.

## Checklist

- [ ] `make build` — not run; needs the generated `web/build` embed assets. `go build ./...` passes.
- [ ] `make test` — both touched packages show the same 6+2 pre-existing Windows failures as clean `main` under an identical sandboxed run (6 credential/HOME in `internal/cli`, `TestWriteFileAtomic_WritesContent` + `TestHeadlessSetupNoTokenFileOK` in `cmd/pad`).
- [x] New features have tests
- [x] TypeScript types updated — N/A, CLI-only
- [x] CLI help text updated — the `--format` flag help

Closes #898. Closes #1076.

### DIFF
--- a/cmd/pad/cmd_attachment.go
+++ b/cmd/pad/cmd_attachment.go
@@ -528,8 +528,13 @@ resolves it to a UUID before calling the API.`,
 				return nil
 			}
 
+			markdown := formatFlag == "markdown"
+			var mdRows [][]string
+
 			tw := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-			fmt.Fprintln(tw, "ID\tMIME\tSIZE\tFILENAME\tITEM\tCREATED")
+			if !markdown {
+				fmt.Fprintln(tw, "ID\tMIME\tSIZE\tFILENAME\tITEM\tCREATED")
+			}
 			for _, raw := range resp.Attachments {
 				var row struct {
 					ID             string  `json:"id"`
@@ -559,9 +564,25 @@ resolves it to a UUID before calling the API.`,
 				if t, err := time.Parse(time.RFC3339, row.CreatedAt); err == nil {
 					created = t.Format("2006-01-02 15:04")
 				}
+				if markdown {
+					mdRows = append(mdRows, []string{
+						short, row.MimeType, humanSize(row.SizeBytes), row.Filename, item, created,
+					})
+					continue
+				}
 				fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t%s\t%s\n",
 					short, row.MimeType, humanSize(row.SizeBytes), row.Filename, item, created)
 			}
+
+			if markdown {
+				cli.RenderMarkdownTable(os.Stdout,
+					[]string{"ID", "MIME", "Size", "Filename", "Item", "Created"}, mdRows)
+				// Pagination footer as prose, so it can't be mistaken for a row.
+				fmt.Printf("\n_%d of %d (limit %d, offset %d)_\n",
+					len(resp.Attachments), resp.Total, resp.Limit, resp.Offset)
+				return nil
+			}
+
 			tw.Flush()
 			fmt.Printf("\n%d of %d (limit %d, offset %d)\n", len(resp.Attachments), resp.Total, resp.Limit, resp.Offset)
 			return nil
--- a/cmd/pad/cmd_item.go
+++ b/cmd/pad/cmd_item.go
@@ -2320,9 +2320,43 @@ func commentsCmd() *cobra.Command {
 				return cli.PrintJSON(comments)
 			}
 
+			if formatFlag == "markdown" {
+				renderCommentsMarkdown(os.Stdout, comments)
+				return nil
+			}
+
 			cli.PrintCommentTable(comments)
 			return nil
 		},
+	}
+}
+
+// renderCommentsMarkdown is the markdown counterpart of cli.PrintCommentTable.
+// Comments are not tabular — the terminal form prints an attribution line then
+// the body — so the markdown form keeps that shape rather than forcing a table.
+//
+// The body is written VERBATIM, not escaped: a comment body is authored as
+// markdown, so escaping it would turn its lists and code fences into literal
+// text. Only the attribution line, which we construct, is sanitized.
+func renderCommentsMarkdown(w io.Writer, comments []models.Comment) {
+	if len(comments) == 0 {
+		fmt.Fprintln(w, "No comments.")
+		return
+	}
+
+	for i, c := range comments {
+		badge := c.CreatedBy
+		if c.Author != "" && c.Author != c.CreatedBy {
+			badge = c.Author + " (" + c.CreatedBy + ")"
+		}
+		fmt.Fprintf(w, "**%s** · %s via %s\n\n",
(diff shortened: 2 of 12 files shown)


xarmian (approved):
Thanks for this — and apologies for the slow turnaround; this deserved a faster response.

Reviewed and verified beyond CI (which last ran on the Aug-12 base):

- Rebuilt and ran the full test suite + lint on this branch **merged onto today's main** — green (main moved 4 PRs since your base; no textual or semantic conflicts).
- Ran the seven new surfaces against live workspaces — `item deps` sections, `item comments` attribution + verbatim body, and the empty states all behave exactly as the PR body documents.
- Probed `ansiPattern` against unterminated CSI/OSC at end-of-string: degrades to stripping the bare ESC with only printable residue left, never leaked control bytes. The test table (bare ESC, ST-terminated OSC, private-mode CSI) is thorough.
- Your two flagged judgment calls both hold up: verbatim comment bodies mirror the existing terminal path (bodies were never escaped there either), and the `displayWidth` broadening fixes a same-family alignment bug — glad you kept it in.

This also closes our internal tracking of the ANSI-stripping gap (BUG-2503). Merging.

PR #1086, opened 2026-08-13
fix(store): include the cursor's own second in /changes deltas (BUG-2539)

xarmian:
Fixes BUG-2539.

## The defect

`items.updated_at` / `items.deleted_at` are TEXT columns holding RFC3339 **whole-second** strings (`store.now()`). The `/changes` cursor is a unix-**millisecond** value — normally the previous response's `server_time`. `ItemsModifiedSince` formatted that cursor with the same second precision, truncating it DOWN, then compared with a strict `>`. Every change landing in the cursor's own second compared equal and was dropped — **permanently**, because the caller then advances its cursor to `server_time` and nothing reaches back.

User-visible symptom: bulk-archive an item ~450ms after a page seeded its cursor and the page keeps rendering it as LIVE indefinitely — no banner, no redirect — while the server has `deleted_at` set.

It was never archive-specific. Updates were dropped identically; a missed update is usually re-delivered by the next event, a missed deletion never is.

## Evidence for the mechanism

Four cursors against a live instance, **all strictly earlier than the archive in real time**, so a correct implementation returns it in all four:

```
archive at epoch_ms 1786630404453 (453ms into its second)
  since = sec_start +   1ms  (same second)      -> MISS   updated=0
  since = sec_start + 400ms  (same second)      -> MISS   updated=0
  since = sec_start -   1ms  (previous second)  -> HIT    updated=9
  since = sec_start -1001ms  (two secs earlier) -> HIT    updated=9
```

Two cursors 1ms and 400ms into the same second miss; one 1ms earlier hits. That is truncation, not a race.

## The fix

Truncate the cursor to its second explicitly and compare **inclusively** on all three comparisons.

This makes the endpoint at-least-once at the second boundary — rows written inside the cursor's second come back again, on every sync whose cursor stays inside that second. That is safe here because the payload is server **state**, not a delta to apply on top of what the caller has: every in-tree consumer overwrites its row from it, so a repeat is a no-op. The code comment says so, and says that a future consumer treating these rows as increments would not be safe.

Sub-second storage is the other fix and is a migration, not a one-liner: these comparisons are lexicographic on TEXT, and mixing precisions inverts them (`…20.451Z` sorts BEFORE `…20Z`), so a partial move to nanosecond timestamps would break this more quietly than it fixes it.

## Tests

Both are mutation-tested against a genuinely reverted query, and both carry a control leg that passes either way — so what is being measured is the fix, not the harness.

- **`TestItemsModifiedSince_SameSecondCursor`** — same-second leg + previous-second control. Alignment is checked against the timestamps the writes actually STORED (not a wall-clock reading taken before them) and a misaligned attempt is retried, so the leg cannot silently stop testing anything. Fails 3/3 unfixed with all three assertions firing; passes 3/3 fixed.
- **`web/e2e/bug-2539-sync-window.spec.ts`** — the user-visible end. Each attempt proves it hit the mechanism it is named after by comparing two server values (the seed `/changes` `server_time`, which is what the client keeps as `lastSyncTime`, against the stored `deleted_at`) and retries if it did not. The banner only counts if the page read a LIVE row first and the deletion arrived on a `/changes` that was neither the cursor seed nor issued before the archive. Same-second leg fails 4/4 unfixed; control passes 4/4.

## Gates

- `go test ./...` — pass
- `make test-pg` (full suite on Postgres) — pass. The columns are TEXT on both engines, so the defect and the fix behave identically; worth running because SQLite-green has hidden PG bugs here before.
- `npm run check` — 0 errors; web unit suite — 1511 passed
- Full desktop e2e — 159 passed, 0 failed. New spec also green on `mobile-chromium`.
- Codex CLI review to CLEAN (six passes; no P1 at any point — every finding was on whether the tests actually measure the fix, and each round tightened them)

## Not fixed here, and worth its own item

A mutation landing between the page's initial item read and its SSE subscription (~30ms) is delivered to nobody and triggers no sync, so the page never learns. `pendingSyncOnConnect` exists for exactly that shape but is armed only on leader promotion, not on first connect. Separate defect, deliberately out of scope.

Not exercised: the BroadcastChannel leader/peer fan-out path (single tab only), and `/items-changes`, which is seq-based and should not share this defect — but I did not test it and am not claiming it is clean.

https://claude.ai/code/session_01QGbUKZBAZoWdEgiTNWsXag

### DIFF
--- a/internal/store/items.go
+++ b/internal/store/items.go
@@ -4784,10 +4784,14 @@ func scanItems(rows *sql.Rows) ([]models.Item, error) {
 // second and no later query reaches back for it: a bulk archive ~450ms after a
 // page seeded its cursor left the item rendering as live indefinitely.
 //
-// `>=` against the truncated second may re-deliver changes from the boundary
-// second that the caller already has. That is the cheap direction to be wrong
-// in — every consumer of this endpoint applies server state idempotently — and
-// it is bounded to one second of changes per sync. Sub-second storage would be
+// `>=` against the truncated second makes the endpoint AT-LEAST-ONCE at the
+// second boundary: rows written inside the cursor's second come back again, and
+// come back on every sync whose cursor stays inside that second, so a burst of
+// rapid syncs can repeat them more than once. That is the cheap direction to be
+// wrong in, because the payload is server STATE rather than a delta to apply on
+// top of what the caller has — every in-tree consumer overwrites its row from
+// it, so a repeat is a no-op rather than a double-apply. A future consumer that
+// treats these rows as increments would not be safe here. Sub-second storage would be
 // the other fix, but items timestamps are second-precision throughout and
 // mixing formats in one column breaks the lexicographic ordering these string
 // comparisons rely on ("…20.451Z" sorts BEFORE "…20Z"), so precision is a
--- a/internal/store/items.go
+++ b/internal/store/items.go
@@ -4770,11 +4770,30 @@ func scanItems(rows *sql.Rows) ([]models.Item, error) {
 // items that were deleted (hard-deleted or archived) since the timestamp.
 //
 // The updated list includes both active AND recently archived items (those with
-// deleted_at > since). This lets the frontend update archived views correctly —
-// an item that was just archived needs its full data to appear in archived views,
-// not just its ID in the deleted list.
+// deleted_at at/after since). This lets the frontend update archived views
+// correctly — an item that was just archived needs its full data to appear in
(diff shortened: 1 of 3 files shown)


xarmian:
Marked ready on Dave's approval, given verbally in-session (day-33, 2026-08-13): "the PR from wren is fine to approve." Flip executed by the lead as bookkeeping of that decision. Merge follows once the two pending Go jobs finish.

PR #1087, opened 2026-08-13
docs(skill): port the applicable plugin-review corrections into the embed source (TASK-2537)

xarmian:
Closes TASK-2537.

`skills/pad/SKILL.md` is the `//go:embed` source that `pad agent install` writes into user projects. It had drifted from corrections made to the plugin's copy during TASK-2534's review rounds. **Selective port** — the two files diverge by design (surface-agnostic vs Claude-Code-only), so none of the plugin's Claude-Code-specific material comes across.

Every item re-verified against the code, since wording and line numbers differ between the copies.

## Ported, as listed on the task

| Was | Verified against | Now |
|---|---|---|
| Ideation example passed `--content "..." --stdin` together | `cmd_item.go:141` — `--stdin` OVERWRITES the `--content` value, then blocks on `io.ReadAll` with nothing piped | `--stdin` dropped |
| Retro's plan-status flip had no `--comment` | contradicts Key Principle 2, four lines below it | `--comment` added |
| Key Principle 2 named `blocked` as a task status | `templates.go:157` — open / in-progress / done / cancelled | `cancelled` |
| Role-board pointers claimed `pad server open` → `/{workspace}/roles` | `cmd_server.go:999` appends only the workspace slug; no sub-path exists | navigate-to-the-Roles-page wording |

## Also ported — same review rounds, not on the task, found by diffing the copies

- **Eight body-loading `pad item list` calls lacked `--full`.** Since the v0.9 summary shape, `pad item list --format json` returns `cli.ToItemSummaries`, which has **no `content` field at all** — so a skill instructed to "follow ALL returned conventions" was reading titles. That is the whole convention-loading mechanism, silently empty. Fixed in the seven trigger-load examples, the `convention_index` bullet's own example, and the retro's task load.
- **`open "$IMG"` is macOS-only**, in a file installed on every platform.

## Not ported, filed instead

The whoami-gated `pad workspace init` routing guards a blind self-heal in the plugin's bootstrap-failure branch — and this file has no such branch, so there was nothing to correct, only something to write. **BUG-2541** carries the structured version (the two stderr signatures as named cases, the Onboarding entry's missing precondition).

What DID ship here is the hazard itself, four surface-agnostic sentences: bootstrap failing usually means setup; the documented individual-call fallback won't help since it needs the same link; don't run `pad workspace init` blind; gate on `pad auth whoami`; otherwise hand back to the user. Codex pushed on the deferral and was right that leaving the file one step from the hazard wasn't defensible even with the follow-up filed.

## One correction I owe

The warning first said `pad workspace init` "hangs indefinitely, no timeout" — inherited from TASK-2534's write-up rather than checked, and wrong. First-admin setup polls under a 20-minute cap (`bootstrap.go::bootstrapPollTimeout`); the unauthenticated branch polls a CLI auth session that expires after 5 minutes (`cli_auth_sessions.go`). Bounded. The hazard is still real — it blocks an agent's tool call for minutes on a flow only a human at a browser can finish — and the text now says that. Corrected on BUG-2541 too, so that item isn't left asserting it.

## Verification

Through the real embed path, not the diff: rebuilt and ran `pad agent install` for both targets (`claude` → `.claude/skills/pad/SKILL.md`, `codex/cursor/windsurf/opencode` → `.agents/skills/pad/SKILL.md`) into scratch dirs, and confirmed the installed artifacts carry every correction and none of the superseded strings — 9 body-load calls with `--full`, 0 hits for `or blocked, use` / `/{workspace}/roles` / `--content "..." --stdin`.

`go test ./...` green. Codex CLI review to CLEAN (five passes; the last two were mine to fix, not the port's). Rebased onto e91c4fc2.

https://claude.ai/code/session_01QGbUKZBAZoWdEgiTNWsXag


xarmian:
Marked ready by the lead under Dave's delegation (day-33, in-session): draft flips no longer require his per-PR approval when the lead has reviewed the change and finds it ready. Review basis here: diff read in full — single file, scope-check clean (no plugin-only content ported), nine --full additions verified, hazard note carries the corrected bounded-timeout claim. Merge follows CI.

PR #1091, opened 2026-08-14
feat(sessions): session-presence registry + GET /api/v1/sessions (PLAN-2558 S1)

xarmian:
Slice 1 of **PLAN-2558** (IDEA-2544 Phase 3 — web-UI push). Unblocks S2/S3; S3 is the demo money-shot PLAN-2469's campaign is waiting on.

## Why

`pad push` (Phase 1, `da6ce642`) is fire-and-forget with no "no session connected" warning. That is a defensible contract for a CLI verb typed by someone who knows whether their own session is running. It is **not** defensible for a web-UI button: "Push to Claude" that silently goes nowhere is *worse* than the clipboard ferry it replaces, because the user cannot tell the two outcomes apart. Presence lets the UI answer the question before the click, and — once sessions carry a label (S2) — turns the same data into the target picker S5 needs.

It also closes the substrate half of **PLAN-2469 Phase 3** ("presence surface: SessionStart hook → live-sessions view", IDEA-2464). The two Phase 3s turned out to be the same work.

## What

- `internal/server/session_presence.go` — `SessionPresence` interface + `MemorySessionPresence`.
- `handleWatchEventsStream` registers on connect, deregisters in a defer **paired with `Unsubscribe`'s on the adjacent line**, so presence is bracketed to the subscription's lifetime and every exit path (`ctx.Done`, a failed SSE write, the replay-loop returns, the reval-tick paths) releases both or neither. A leaked entry is the failure that matters — it makes the UI promise a listener that is gone.
- `internal/server/handlers_sessions.go` — `GET /api/v1/sessions`.

## Two decisions worth a reviewer's eye

**Self-scoped, no admin view.** No `?user_id=`, no admin bypass, unlike most list endpoints here. Who has an agent session open is a presence signal about a person, not a fact about workspace content — the same reasoning that made push self-addressed only. A cross-user view is a product decision with a consent design attached, not a query parameter.

**503, not 200-with-an-empty-list**, when no registry is wired. "I can't tell" and "nobody is listening" must not look the same to the UI; collapsing them is precisely the dishonesty this slice exists to remove.

## Multi-process boundary

`MemorySessionPresence` is per-process, so it is an interface from day one. The doc comment states the boundary precisely rather than hand-waving it: `watchevents.Bus` is blind in the **same** direction (a push published on instance A never reaches a stream on instance B), so per-instance presence is as accurate as per-instance delivery and both stop being trustworthy at the same boundary — *except* that a load balancer may route a POST and a GET to different instances, at which point the two disagree. A Redis-backed presence must therefore land **with** the Redis `watchevents.Bus` that package already anticipates, not separately. Fixing one alone just moves the disagreement.

**Do not put the web-UI push surface (S3) in front of a multi-process deployment until both exist.**

## pad-cloud

No change required — checked, not assumed. `/api/v1/sessions` is a plain JSON GET served by `nginx-router.conf`'s default `location /` pass-through; the special long-lived-connection blocks are for `/api/v1/events` and `/api/v1/collab/` only.

## Verification

- `make test` (SQLite) clean
- `make test-pg` clean — 25 packages, exit 0
- `make lint` — 0 issues
- New tests pass under `-race` (including a concurrent Add/List/Remove hammer)
- **Live, on the installed binary:** 0 sessions with nothing connected → 1 with one stream open → 2 with two, oldest-first → back to 0 after both disconnect, with a second user's list staying empty throughout.

### DIFF
--- a/cmd/pad/cmd_server.go
+++ b/cmd/pad/cmd_server.go
@@ -662,6 +662,17 @@ func serveCmd() *cobra.Command {
 			// eventBus above, which branches on PAD_REDIS_URL).
 			srv.SetWatchEventsBus(watchevents.New())
 
+			// Live-session presence registry (PLAN-2558 S1). Wired
+			// unconditionally for the same reason and with the same
+			// caveat as the watch bus directly above: in-process only,
+			// no Redis-backed implementation yet. It reports on the
+			// same connections that bus feeds, so the two share a
+			// lifetime and a limitation — see
+			// internal/server/session_presence.go's SINGLE-PROCESS
+			// LIMITATION note before putting the web-UI push surface
+			// (PLAN-2558 S3) in front of a multi-process deployment.
+			srv.SetSessionPresence(server.NewMemorySessionPresence())
+
 			// Yjs collab room manager (PLAN-1248). Single-instance only
 			// today; multi-replica fanout via Redis is a deferred IDEA.
 			// MemoryOpBus is in-process; the OpBus interface keeps the
--- a/internal/server/handlers_sessions.go
+++ b/internal/server/handlers_sessions.go
@@ -1,0 +1,55 @@
+package server
+
+import "net/http"
+
+// sessionsResponse is the body of GET /api/v1/sessions.
+//
+// Count is redundant with len(Sessions) and is included anyway: the
+// primary consumer is a UI that only needs "is anything listening?" to
+// decide whether the push button is honest (PLAN-2558 S3), and making
+// that the cheapest possible read keeps the common case from having to
+// reason about the array at all.
+type sessionsResponse struct {
+	Sessions []LiveSession `json:"sessions"`
+	Count    int           `json:"count"`
+}
+
+// handleListSessions returns the caller's OWN live event-stream
+// connections (PLAN-2558 S1). It is the read side of the presence
+// registry described in session_presence.go.
+//
+// SELF-SCOPED, WITH NO ADMIN VIEW. There is deliberately no
+// ?user_id= parameter and no admin bypass, even though most list
+// endpoints in this server have one. Who has an agent session open, and
+// when, is a presence signal about a person rather than a fact about
+// workspace content — the same reasoning that made `pad push`
+// self-addressed only (handlers_push.go: "pushing into someone else's
+// session is a consent question, not a code question"). If a
+// cross-user view is ever wanted, it is a product decision with a
+// consent design attached, not a query parameter.
+//
+// GET /api/v1/sessions
+func (s *Server) handleListSessions(w http.ResponseWriter, r *http.Request) {
+	if s.sessionPresence == nil {
+		// Mirrors handleWatchEventsStream's own nil-bus stance: a server
+		// built without the presence registry doesn't have a degraded
+		// answer to give, and answering 200 with an empty list would be
+		// a LIE in exactly the direction this feature exists to prevent
+		// (the UI would report "nobody is listening" when it simply
+		// cannot tell).
+		writeError(w, http.StatusServiceUnavailable, "unavailable", "Session presence is not available")
+		return
+	}
+
+	user := currentUser(r)
+	if user == nil {
+		// Same stance as the stream this endpoint reports on: there is
+		// no coherent "my sessions" for a workspace-scoped token or the
+		// fresh-install no-auth window. Resolved user identity or 401.
+		writeError(w, http.StatusUnauthorized, "unauthorized", "Authentication required")
+		return
+	}
+
+	sessions := s.sessionPresence.ListForUser(user.ID)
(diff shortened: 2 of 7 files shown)


xarmian:
## CI: the `Go` job red here is pre-existing on main, not this PR

Verified rather than assumed:

- Failing step is **`Run govulncheck`** — not build, lint, or test. `make test`, `make test-pg` and `make lint` are clean locally on this branch; CI's Web and both Smoke jobs pass.
- **main is red on the same step** at `da6ce642`, before this branch existed: `f9195c5b` (21:15Z) green → `da6ce642` (22:41Z) red, on a commit touching no dependencies and no toolchain pin. Most likely a vuln-DB entry published in that window.
- Cause: 8 Go **standard library** vulns, all `Found in: net@go1.26.5 / Fixed in: net@go1.26.6`. No dependency bump fixes it — the toolchain pin has to move.

Filed as **BUG-2565** (high). Proceeding per PLAYB-40 §6 (document pre-existing CI failures and continue). Judge this PR on the Go test steps, Web and Smoke.

## Correction to a claim in the description

The description says the defer arrangement means every exit path "releases both or neither." That's accurate about the *handler's* exit paths, but it undersells one thing, so stating it plainly:

A **clean** disconnect deregisters immediately — `ctx.Done` fires and the defer runs (this is the leg verified live). An **ungraceful** disconnect (half-open TCP, closed laptop, dropped network) isn't noticed until the next keepalive write fails, and `watchEventsKeepaliveInterval` is **30s**. So the list can over-report a dead session for up to ~30 seconds.

That bound is acceptable — push is fire-and-forget either way, so a stale entry costs a message that would have been lost regardless. But it constrains what S3's UI may claim: **"one session connected" is honest; "this will be delivered" is not.** Recorded on TASK-2561 so S3 doesn't inherit the overclaim.

xarmian:
## Review record — converged at round 3

Three codex rounds with framing rotation (PATTE-37), reviewed at `main...a2366b65`.

| Round | Framing | Result |
|---|---|---|
| 1 | "Review this diff" | CLEAN — traced the route tree and confirmed `/api/v1/sessions` inherits the full middleware stack (`TokenAuth, SessionAuth, RateLimit, CSRFProtect, RequireAuth`) despite being registered outside the nested `/api/v1` block |
| 2 | "What might round 1 have missed?" | 2 × P2 — both real, both verified against source before acting |
| 3 | "Contract-change blast radius" | CLEAN — no existing caller depended on old semantics; stream wire/auth/reconnect behavior unchanged, presence is additive and nil-safe |

### Round 2 disposition

**Cache-Control — fixed** (`a2366b65`). Confirmed `writeJSON` sets none and `jsonContentType` sets only Content-Type, so the response was heuristically cacheable. Now `private, no-store`, pinned by a test.

Worth noting the header is belt-and-braces at the proxy layer, not load-bearing there: neither `deploy/nginx.conf` nor pad-cloud's `nginx-router.conf` defines a `proxy_cache_path`, so nginx caches nothing regardless. The header is for browsers and any intermediary we don't control.

**Shutdown — refined, not adopted as reported.** The mechanism is correct: `Server.Shutdown` delegates to `http.Server.Shutdown`, which does not cancel in-flight handler contexts, so SSE handlers hang until their own `ctx.Done` or a failed write. But for `MemorySessionPresence` this is **harmless** — the registry lives in the process that is going away, so entries die with it and there is nothing to reap. The finding's value is forward-looking, and sharper than reported: a Redis-backed implementation does not inherit that, since its entries outlive the writing process and a crash strands them *permanently* rather than for 30 seconds. Recorded as a hard constraint on the `SessionPresence` interface.

### Calibration note

The 30s staleness window documented in `a2366b65` was found by hand, **not** by codex. The orchestrator confirmed codex read the exact keepalive machinery — the interval, the ticker, the "keepalive write failed, closing" line — across two rounds and did not connect it to a presence finding. It was in view and missed, twice. Recording that because it is a cleaner demonstration than usual that a CLEAN round is not proof of absence.

### Open property, examined and accepted

No cap on presence entries per user. With a ~30s reap window, a flapping client reaches a steady state of roughly (reconnect rate × 30) entries rather than growing without bound, each entry is a few dozen bytes, and the endpoint sits behind auth plus the general 600/min per-user limiter. Accepted rather than bounded — flagged here so it is a decision on the record, not an oversight.

### Verification on `a2366b65` (the amended branch, not just the original commit)

- `make test` — exit 0, 25 packages, 0 failures
- `make test-pg` — exit 0, 25 packages, 0 failures
- `make lint` — 0 issues
- Presence tests green under `-race`
- Live legs re-confirmed earlier on the installed binary: 0 → 1 → 2 (oldest-first) → 0, second user's list empty throughout

CI's `Go` job will still show red on `Run govulncheck` — pre-existing on main, BUG-2565, see the comment above.

PR #1102, opened 2026-08-15
test(server): drive the reval-fault test off the fault seam instead of a sleep (BUG-2570)

xarmian:
## Summary
Fixes the watch-stream CI flake (BUG-2570, two instances on 2026-08-14/15, both timing out at the control-event wait). The reload-fault closure in `TestWatchEventsStream_RevalStillDeniesWhenWatchListReloadErrors` is now the test's clock: it narrows the member's access on faulting tick 1 and lifts the fault on faulting tick 2, so consecutive reload failures stop at exactly 2 — strictly below `maxConsecutiveWatchReloadFailures` — and the clear-the-watch-set path can never silence the control leg. The 300ms sleep is gone.

## Mechanism of the flake
Within a reval tick the handler runs `visCache.reset()` before the reload, so the closure fires strictly after its own tick's reset, and ticks are sequential in the handler goroutine. The old shape (install fault + narrow access + sleep 300ms at a 200ms interval) bet the sleep against both "at least one tick fired" and "fewer than three fired before the control event lands". Loaded runners (~45-50ms/request under race detector) lost the second bet: the third faulting tick cleared the watch set and the control comment never delivered.

## Plus
- Codex round 1 on this fix surfaced that regression *detection* still has a window — a successful tick 3 masks a hypothetical reset-skipped-on-fault regression. Interval set to 500ms (~10x headroom for the one request that must beat tick 3) and the asymmetry documented in place: losing that race can only mask a not-yet-written bug in one run, never fail a green build.
- Control-leg wait widened 3s → 10s: it asserts delivery-at-all, not latency, and it is the exact wait that timed out in both CI instances.

## Test plan
- [x] 5x `-race` green at 50ms and 500ms intervals
- [x] Counterfactual: mutant with `visCache.reset()` moved to the reload success path leaks 3/3 runs (test detects the guarded regression deterministically at 500ms)
- [x] Full `go test ./...` + `make lint` green; Postgres leg (`TestWatchEventsStream|TestWatchVisCache`) 2x `-race` green
- [x] Codex rounds: R1 found the tick-3 masking window (addressed); R2 (rotated lens: cross-test interference, cleanup ordering, package timeout budget) CLEAN

Refs: BUG-2570

https://claude.ai/code/session_01BhQoeaWXxJbvw86ezzK8dt

### DIFF
--- a/internal/server/handlers_watch_events.go
+++ b/internal/server/handlers_watch_events.go
@@ -198,6 +198,13 @@ func (s *Server) handleWatchEventsStream(w http.ResponseWriter, r *http.Request)
 	defer keepalive.Stop()
 	reval := time.NewTicker(watchListRevalInterval)
 	defer reval.Stop()
+	// Test seam (BUG-2570): a test may substitute its own tick source so
+	// each revalidation tick happens exactly when the test says, and never
+	// otherwise. Read once at setup — see watchRevalTickOverride's doc.
+	revalC := reval.C
+	if ch := s.watchRevalTickOverride.Load(); ch != nil {
+		revalC = *ch
+	}
 
 	ctx := r.Context()
 	for {
@@ -225,7 +232,7 @@ func (s *Server) handleWatchEventsStream(w http.ResponseWriter, r *http.Request)
 			}
 			flusher.Flush()
 
-		case <-reval.C:
+		case <-revalC:
 			// TASK-2533 codex round 4: reset the identity/visibility
 			// cache FIRST, unconditionally, BEFORE attempting the watch-
 			// list reload — the two used to be coupled (visCache.reset()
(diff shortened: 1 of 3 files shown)


xarmian:
Superseding the interval-tuning approach after codex rounds: R1 (two independent runs) surfaced two regression-detection windows a free-running ticker can't close — a stray successful tick before fault installation (resets visCache pre-fault) or after the tick-2 lift (reset + access-filtered reload) masks the reset-skipped-on-fault regression. Interval tuning trades green-determinism against detection-determinism; the seam gives both. New commit adds watchRevalTickOverride (mirrors the watchPredicatesLoadFault seam pattern; nil in production, one extra atomic load at stream setup) and the test now drives exactly one reval tick, post-narrow, fault-active. Verified on the final shape: 5x+3x -race green, counterfactual mutant leaks 5/5 deterministically, full suite + lint green, PG leg 2x -race green, codex blast-radius round CLEAN.

PR #1120, opened 2026-08-16
fix(server): serve RFC 9728 PRM at path-aware well-known (BUG-2266)

xarmian:
## Summary

Serves the RFC 9728 protected-resource metadata document at the path-aware well-known URLs (`/.well-known/oauth-protected-resource/mcp` and trailing-slash `/mcp/`) in addition to the existing exact-match root. Fixes OAuth discovery for MCP clients configured with the path-suffixed transport URL (`https://mcp.getpad.dev/mcp` — the shape every FastMCP example uses, e.g. Kimi CLI): per RFC 9728 §3.1 such clients insert the well-known segment *before* the path, and that URL previously fell through to the SPA catch-all, so discovery died JSON-parsing an HTML shell.

Bounded to the two real shapes rather than a wildcard: the handler emits `Cache-Control: public, max-age=3600`, and a wildcard would hand a CDN one cacheable object per attacker-chosen suffix (codex round 2 finding, fixed).

Scope: PRIMARY fix from BUG-2266 only. The audience-widening secondary (`NormalizeAudience` / `audienceMatchingStrategy`) is deliberately untouched — separate security-boundary item. For the same reason the suffixed doc keeps the canonical bare-host `resource`: echoing `.../mcp` would steer compliant clients into an audience the AS still rejects (codex round 1 P1, declined with reasoning — doc-following clients converge on the canonical audience and work end-to-end; strict §3.3 validators fail cleanly at discovery exactly as before, until the audience item lands).

## Test plan

- New `TestMCP_DiscoveryDoc_PathAwareWellKnown`: decodes `/mcp` and `/mcp/` variants into the typed doc and compares field-by-field against the root response (SPA HTML or a 404 body cannot satisfy it); pins that an arbitrary suffix (`/bogus`) does NOT get the doc. `t.Parallel` per CONVE-2086.
- Path-aware URL added to `TestMCP_CloudModeOff_RoutesAbsent`'s 404 list (cloud-mode gate preserved).
- Mutation-verified: with the route lines removed, the new test fails `expected 200, got 404`.
- Live-verified on the real binary (cloud-mode boot, temp HOME/DB): suffixed paths return `200 application/json` with the full doc; root unchanged; control leg confirms an unmatched well-known path still serves the SPA shell — the pre-fix behavior at the suffixed path.
- Codex: round 1 FINDINGS (P1 declined on coupling grounds, recorded above), round 2 fresh-angle sweep — middleware/auth ordering, WWW-Authenticate pointer consistency, route shadowing all CLEAN; cache P2 fixed by bounding. Round 3 confirm pending below.

Closes BUG-2266.

### DIFF
--- a/internal/server/handlers_mcp.go
+++ b/internal/server/handlers_mcp.go
@@ -174,5 +174,19 @@ func (s *Server) registerMCPRoutes(r chi.Router) {
 	// (protected-resource) gets the real metadata; RFC 8414 (auth-server)
 	// is the 501 stub TASK-951 fills in.
 	r.With(s.requireCloudMode).Get("/.well-known/oauth-protected-resource", s.handleOAuthProtectedResource)
+	// Path-aware RFC 9728 §3.1 variant (BUG-2266). A client configured
+	// with the path-suffixed transport URL (https://mcp.getpad.dev/mcp —
+	// the shape every FastMCP example uses) constructs its metadata URL
+	// by inserting the well-known segment BEFORE the path:
+	// /.well-known/oauth-protected-resource/mcp. Without this route the
+	// request fell through to the SPA catch-all and discovery died
+	// JSON-parsing HTML. Bounded to the two shapes a pasted transport
+	// URL actually produces (/mcp and a trailing-slash /mcp/) rather
+	// than a wildcard: the handler sets Cache-Control public max-age,
+	// and a wildcard would hand a CDN one cacheable object per
+	// attacker-chosen suffix (codex round 2). /mcp is the only
+	// path-mounted resource, so nothing else needs the variant.
+	r.With(s.requireCloudMode).Get("/.well-known/oauth-protected-resource/mcp", s.handleOAuthProtectedResource)
+	r.With(s.requireCloudMode).Get("/.well-known/oauth-protected-resource/mcp/", s.handleOAuthProtectedResource)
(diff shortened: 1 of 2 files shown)


xarmian:
LEAD GO (review record). Verified independently: diff is 2 files as claimed, audience.go untouched — scope held to the PRIMARY fix per the BUG-2266 re-grounding pre-commit. The bounded /mcp + /mcp/ routes (over my allowed wildcard) are the better call given Cache-Control public — the CDN cache-key rationale is documented in-route. Tests verified real, not vacuous: the path-aware test decodes typed JSON field-by-field against the root variant (an SPA shell or 404 body cannot satisfy it), the bogus-suffix negative pins the boundedness, and the per-route requireCloudMode test mounts-then-flips so only the middleware can refuse. CI 7/7 green on 0721e3b2, checked directly.

One residual ON RECORD (accepting Wren's declined P1 as the scoped-correct choice): the suffixed metadata URL serves the CANONICAL bare-host resource value rather than echoing the /mcp-suffixed identifier. If a strict RFC 9728 client validates the returned resource field against its configured transport URL, discovery still fails for it — that failure mode is coupled to the out-of-scope audience-normalization item (see the BUG-2266 trail on why that is a separate, security-boundary-reviewed change). If the original reporter's client (FastMCP/Kimi) still fails after this ships, that residual is the first suspect and the follow-up item exists.

Cleared to flip and squash-merge under the 2026-08-13 standing grant.

PR #1121, opened 2026-08-16
fix(mcp): explicit tool annotations from catalog write-shape knowledge (BUG-2302)

xarmian:
## Summary

Every Pad MCP tool inherited mcp-go's `NewTool` default annotations (`readOnlyHint:false, destructiveHint:true, openWorldHint:true`), so pure reads like `pad_search` advertised themselves as destructive — training users to click through host confirmation prompts.

Annotations are now derived explicitly in `buildToolFromDef` from the catalog's own knowledge, three classes:

- **read** (`pad_search`, `pad_project`, `pad_attachment`, `pad_meta`, `pad_playbook`): `readOnlyHint:true, destructiveHint:false, idempotentHint:true`
- **additive writes** (`pad_workspace` — invite/create/claim/restore; `pad_library` — activate): `readOnlyHint:false, destructiveHint:false` (codex round 1: marking additive writes destructive reintroduces the prompt-training harm)
- **overwrite/delete-capable** (`pad_item`, `pad_collection`, `pad_role`): conservative `destructiveHint:true` — unchanged on the wire from the old defaults

`openWorldHint:false` everywhere (closed-world tools). `pad_set_workspace` gets a hand-written block. `readOnlyActions` stays the single read source; a sibling `additiveWriteActions` allowlist (absence ⇒ destructive) carries the write-shape knowledge. Also fixes the missing `pad_item.history` read-only entry (documented read-only since v0.14).

**ToolSurfaceVersion 0.19 → 0.20** (behavior bump — no tool names, action enums, or param shapes changed). Per the lead's sequencing ruling, the 0.20 changelog entry describes only this change; BUG-2305 appends to it if it ships in the same window (one bump total).

## Test plan

- `TestCatalogTools_AnnotationsExplicit`: literal per-tool read/additive/destructive table — deliberately a second enumeration so a broken derivation, a new tool, or a class-changing action addition fails loudly. Mutation-verified both directions (polarity flip fails 10 tools; always-destructive fails the 2 additive rows).
- `TestAdditiveWriteActions_NoStaleEntries`: allowlist entries must be real catalog pairs and never overlap `readOnlyActions`.
- `TestSetWorkspaceTool_AnnotationsExplicit`: both deployment variants.
- `pad_item.history` joins the read spot-checks in `TestToolSurfaceJSON_ReadOnlyFlags`.
- Gates: build OK · `make lint` 0 issues · `go test ./...` exit 0 · test-pg N/A (no store changes) · web N/A.
- Codex (PLAYB-1832 oracle form): R1 FINDINGS (additive over-classification — fixed), R2 CLEAN on fresh angles (idempotency semantics, wire serialization of explicit false pointers, allowlist drift risk).

Closes BUG-2302.


xarmian:
LEAD GO (review record), with one pre-merge nit. Verified independently: 8-file diff matches the package claim; annotationForDef derives from readOnlyActions/additiveWriteActions as the single source; the test's expectation table is a genuine SECOND enumeration with stale-row checks in both directions and nil-rejection pinning explicitness (the pre-fix defect shape); the history allowlist gap is closed with the serializer expectation updated; the 0.20 changelog entry describes ONLY this PR's change per the sequencing ruling, with the behavior-bump precedent cited; destructive-class tools are wire-identical to the old defaults except OpenWorldHint true→false, which the bump covers. CI 7/7 on b21310a6, checked directly. The additive class rationale (codex R1's over-classification catch) is well-reasoned and documented at the allowlist.

NIT (fix in-branch before merge, no re-review needed): the new annotationForDef doc block was inserted directly after buildToolFromDef's original doc comment, so that comment ('...JSON Schema discriminated unions are awkward... the description carries the action × params matrix anyway') now attaches to annotationForDef and buildToolFromDef is left undocumented. Move the original comment back above buildToolFromDef.

Cleared to flip and squash-merge under the standing grant once the nit lands. BUG-2305 rebases onto the merge and appends to the 0.20 entry as sequenced.

PR #1122, opened 2026-08-16
fix(mcp): summary-shaped item list on the HTTP transport (BUG-2305)

xarmian:
## Summary

A bare `pad_item.list` over remote `/mcp` returned up to 50 items with **full content bodies** — the exact token blowup TASK-2000 was written to prevent, unmitigated on the transport the zero-CLI plugin makes primary. The limit was symmetric; the shape was not: the exec path projects through `cli.ToItemSummaries` in the CLI, the HTTP path forwarded raw handler responses.

Fix follows the trail's pre-committed approach (the body's mapper-param sketch is refuted there): a hand-written `dispatchItemList` — same shape and same projection as `HTTPResourceFetcher.fetchItemList` — with `mapItemList` retained as the single URL/filter builder and the routeTable entry removed. A newly **declared `full` boolean** on `pad_item` is the discoverable opt-in for complete bodies, honored on both transports (stdio forwards it as the CLI's `--full`). `buildRequestErrorResult` is extracted from `executeRequest` so scope denials on the summary path surface as `permission_denied`, not `server_error` (codex R1).

Audit per the body's ask: `pad_search` is symmetric (the store zeroes `Item.Content` in search results); `pad_project.activity` returns enrichment metadata, no bodies. Only item list had the asymmetry. Also rewrites the misleading `actionItemList` comment, corrects scope-test fixtures that stubbed an object shape the endpoint never returns, and forward-points the v0.9 changelog's now-false "not yet surfaced" claim.

**No version bump**: appends to the existing 0.20 changelog entry (BUG-2302 carried the single 0.19→0.20 bump; lead's one-bump sequencing ruling).

## Test plan

- `TestHTTPItemList_DefaultIsSummaryShape` + `TestHTTPItemList_FullOptsIntoCompleteBodies`: real server + store, counterfactual pair — the marker sits past the `content_preview` cut; the full leg proves it flows through the same path, the default leg proves the projection strips it.
- `TestHTTPItemList_ScopeDenialIsPermissionDenied`: pins the error envelope on the summary path (the broken version answers `server_error`).
- Mutation-verified: removing the dispatch case fails (route gone); skipping the projection fails (leak caught).
- Gates: build OK · `make lint` 0 issues · `go test ./...` exit 0 · test-pg N/A · web N/A (README/instructions text only).
- Codex (oracle form): R1 FINDINGS (`full` undiscoverable; permission-envelope bypass — both fixed), R2 FINDINGS (doc drift — fixed), R3 CLEAN.

Closes BUG-2305.


xarmian:
LEAD GO (review record). Verified independently: dispatchItemList follows the trail's pre-commit exactly — mapItemList retained as the sole URL builder (routeTable absence documented in place), summary path projects via cli.ToItemSummaries matching the resources_http.go precedent, full:true forwards through the identical executeRequest path. The buildRequestErrorResult extraction is a pure move serving both call sites, closing the envelope asymmetry codex R1 caught. The stdio leg of the 'full' claim traces: BuildCLIArgs maps input keys generically over the CLI flag tree and 'pad item list --full' exists at cmd_item.go, so the param is honored on both transports as advertised — the v0.18 lesson correctly applied. Changelog appends to 0.20 with no second bump per sequencing, and the v0.9 entry's historical correction keeps its original claim true-as-of-then while noting the v0.20 closure — the right way to amend history.

Test design is the strongest of today's three: real server + store precisely because false-shape stub fixtures ({"items":[]} vs the endpoint's bare array) are what hid this bug; the second-line marker distinguishes full-body leak from legitimate preview; the counterfactual full:true leg proves the missing marker is projection, not an empty fixture; the scope-denial test discriminates the pre-extraction failure mode. CI 7/7 on a5ed254f, checked directly.

Cleared to flip and squash-merge under the standing grant.

PR #1123, opened 2026-08-16
fix(web): full-page pane host gets its own navigate-away handling (BUG-2178)

xarmian:
## Summary

The full-page item host reused the pane controller's collection-host-shaped `handlePaneNavigateAway`, so a collection rename or item move performed from within the docked pane abandoned the master route. The host now wires `planFullPageNavigateAway` — a new pure planner beside `planPaneDrill`:

- **Collection rename** (keeps-pane emit): **ignore**. The master's own BUG-2272 SSE handler self-heals the shared-collection case with `?item=` preserved; a foreign collection never touched the master route; the embedded pane trusts `item.collection_slug`. Known degradation (lost SSE → dead route) is BUG-2272's existing exposure class, documented in the planner comment and routed to **BUG-2601** (delta-sync doesn't recover collection renames).
- **Item move** (no `?item=`): **retarget** the pane to the moved item via the existing drill machinery — preserves the master pathname by construction. The same-slug noop edge self-heals via SSE adoption, which now also refreshes the collection object: new `refreshCollectionIfMoved` (embedded-only, gen-fenced + write-time semantic recheck) at all three adoption sites (SSE, incremental, full-refresh) — fixes stale-schema rendering for remote moves too (codex R1/R2).
- **Malformed URLs**: ignore — the hostile-input unit test caught a `decodeURIComponent` crash pre-ship.

The controller's `handlePaneNavigateAway` behavior is byte-identical for the collection host (comment-only diff there). Pre-existing semantic-fencing gaps at other collection write sites are filed as **BUG-2602** with the class analysis.

## Test plan

- `planFullPageNavigateAway` unit table (both real emit shapes verbatim; encoded slugs; trailing slash; malformed/empty/foreign-origin inputs).
- E2E: move the pane item to another collection from the docked pane on the full-page host — asserts the pathname never leaves the master route and `?item=` retargets. **Mutation-verified against a control binary with the old wiring: fails with the master abandoned, the reported bug verbatim.**
- Gates: `svelte-check` 0 errors · vitest 33/33 (planner) + full suite green pre-amend · `go test ./...` exit 0 · `make lint` 0 issues · web build OK.
- Codex: R1 FINDINGS (2 P1s — one accepted as existing exposure + filed, one folded in), R2 FINDINGS (semantic fence, full-refresh gap, non-embedded mixed state — fixed/filed per disposition), R3 CLEAN (fresh angle: drill depth/ownership arithmetic).

Closes BUG-2178.

### DIFF
--- a/web/e2e/pane-full-page-capstone.spec.ts
+++ b/web/e2e/pane-full-page-capstone.spec.ts
@@ -43,8 +43,12 @@ import type { SuiteFixture } from './fixtures';
  *      the master route).
  *
  * DEFERRED / out of scope (do NOT assert): BUG-2177 (an editor-bound
- * upload/crop orphaned by the peeking editor remount — accepted) and BUG-2178
- * (collection-rename/move-from-pane URL shaping on this host — deferred).
+ * upload/crop orphaned by the peeking editor remount — accepted). BUG-2178
+ * (collection-rename/move-from-pane URL shaping on this host) is now FIXED
+ * and covered below: the move-from-pane case is asserted end-to-end; the
+ * rename case is a host-side IGNORE whose decision table lives in
+ * planFullPageNavigateAway's unit tests (paneController.test.ts) and whose
+ * route self-heal is BUG-2272's separately-tested SSE machinery.
  *
  * The pane is a desktop-split concern, so the desktop project alone covers it.
  */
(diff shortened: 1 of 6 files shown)


xarmian:
LEAD GO (review record). Verified independently: paneHostController.ts is comment-only as claimed — behavior byte-identical for the collection host, exactly the condition this item's re-grounding set. The planner is the right shape: pure, unit-tabled against both real emit shapes verbatim, with the hostile-input floor (the decodeURIComponent catch earned its place — the test found a real crash). The IGNORE-on-rename disposition is properly grounded in BUG-2272's existing exposure with BUG-2601 cross-referenced in BOTH directions (verified), and the retarget path's only action is a drill, which preserves the master pathname by construction. The refreshCollectionIfMoved double-fence (gen + write-time semantic recheck) with its embedded-only guard is well-reasoned and documented at the site; BUG-2602's class analysis of the pre-existing single-fenced sites is the right routing for the discovered class. E2e discriminates on pathname preservation — the exact reported failure shape. CI 7/7 on 66e2e39b, checked directly.

ONE NIT (in-branch before merge, no re-review): the paneHostController.ts annotation says 'planFullPageNavigateAway (below)' — the planner lives in paneController.ts, not below in that file. Say '(paneController.ts)'.

OBSERVATION, no action: a foreign-origin absolute URL without ?item= plans a retarget, but the target only feeds the internal drill machinery as a slug — the safe floor holds; noting so nobody re-derives it.

Cleared to flip and squash-merge under the standing grant. This closes the day-41 restock.

PR #1124, opened 2026-08-16
fix(oauth): keep zero-workspace consent authorizable via the wildcard path (BUG-2303)

xarmian:
## Root cause

The consent template gated the entire workspace fieldset on `.Workspaces` being non-empty. For a user with zero memberships no `workspace_access` radio rendered at all, so the inline script's initial paint hit the missing-radio branch and permanently disabled Authorize — a hard dead end. The server was never the blocker: `parseConsentPayload`'s wildcard path accepts a zero-workspace `workspace_access=all` consent with no membership validation, and `may_create_workspaces` defaults on.

Zero-workspace users are rarer than the bug body claimed (all three cloud signup paths auto-create a workspace; the state is reached via auto-create failure or deleting/leaving your only workspace — severity re-rated on the item), but the state is a real dead end and gates the zero-CLI funnel (PLAN-2309): the fixed page is what lets an MCP client mint a token that can create the user's first workspace.

## Change

- Render the "All my workspaces" radio unconditionally; force-checked when memberships are zero (it is the only option — an unchecked radio group would re-disable the button).
- Keep the "Only specific workspaces" radio + picker gated on `.Workspaces` (nothing to pick).
- Replace the dead-end red copy with helper copy pointing at the workspace-creation checkbox; drop the now-unused `.no-ws` CSS.
- Keep the JS missing-radio branch as a fail-closed guard (comment updated).

Template-only; no server logic changed.

## Tests

- `TestConsent_NoWorkspaces_ShowsEmptyState` (which pinned the buggy dead-end) rewritten as `TestConsent_NoWorkspaces_WildcardStaysAuthorizable`: pre-checked wildcard radio, no specific radio/picker, dead-end copy gone, creation checkbox checked. Assertions locate the `<input>` tag then check attributes inside it (Codex round-2 finding: attribute-order brittleness).
- New `TestConsent_ApproveWildcard_ZeroWorkspaces`: new-shape approve (`workspace_access=all` + `may_create_workspaces`) from a zero-workspace user mints a code and the connection row carries AllCurrent + IncludeFuture + MayCreate.

## Verification

- Negative control: render test against the pre-fix template (via `git show HEAD~1:`) fails on exactly the fixed behaviors; approve test passes pre-fix, confirming the bug was UI-only.
- Browser leg (JS behavior Go tests can't see): rendered the real handler output for a zero-workspace user (control + fixed), drove both in Playwright. Control: Authorize disabled, no radio, form posts NO `workspace_access`. Fixed: Authorize enabled, radio checked, form posts `workspace_access=all` + `may_create_workspaces=1`, zero JS errors.
- `go test ./...` exit 0; `make lint` 0 issues; Codex rounds R1 CLEAN, R2 one P2 (fixed), R3 confirm CLEAN. `make test-pg` result to follow on BUG-2303.

### DIFF
--- a/internal/server/handlers_oauth.go
+++ b/internal/server/handlers_oauth.go
@@ -1426,7 +1426,6 @@ button:disabled { opacity: 0.5; cursor: not-allowed; }
 button.primary:hover:not(:disabled) { background: #1e54d4; }
 .footer { margin-top: 2em; font-size: .85em; color: #888; }
 .footer a { color: inherit; }
-.no-ws { color: #b00; }
 .allow-helper {
   color: #b00; font-size: .85em; margin-top: .35em;
   display: none;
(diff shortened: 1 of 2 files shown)


xarmian:
Lead review (artifact + diff): GO. Scope verified template-only — the single JS hunk keeps the missing-radio branch as a fail-closed guard, which is the right disposition. Checked the wildcard radio's checked-condition against all three membership states (zero → force-checked sole option; non-empty+DefaultWildcard → checked as before; non-empty without → specific stays default): no behavior change outside the zero-workspace state. Tag-scoped test assertions are the right fix for R2's brittleness finding, and TestConsent_ApproveWildcard_ZeroWorkspaces passing pre-fix is the cleanest possible proof of UI-only scope. Verification package (negative control on HEAD~1, Playwright control/fixed legs, 7/7 CI, codex R1/R2/R3 converged) meets the bar. This comment is the on-record approval (same-account precedent, search #17): flip ready and squash-merge under your grant.

PR #1126, opened 2026-08-16
fix(mcp): route backlinks/history/report over HTTP transport + catalog↔route parity test (BUG-2304)

xarmian:
## Root cause

Three catalog actions — `pad_item.backlinks`, `pad_item.history`, `pad_project.report` — were advertised in `tools/list` on both transports but had no HTTP route: no routeTable entry, no special-case dispatcher, no link spec. On Pad Cloud (where HTTP is the MCP transport) every call answered `"not yet implemented over HTTP transport"`. All three have working REST endpoints; the gap was mappers — and nothing enforced catalog↔route parity, which is how they shipped silently (the only prior lock was a hand-picked 12-entry want-list).

The bug's fourth claim (`history` missing from `readOnlyActions`) was already fixed on main by BUG-2302; re-verified before starting.

## Change

- **`item backlinks` + `project report`** — plain routeSpecs. Their CLI `--format json` prints the endpoint response verbatim (`backlinksCmd`, `reportCmd`), so a GET with the declared query params (`limit`/`offset`, `window`/`collections`) reproduces the stdio shape exactly. Both handlers resolve refs and slugs alike.
- **`item history`** — hand-written `dispatchItemHistory`: the versions endpoint returns full `models.Version` rows including every resolved content body, while the CLI projects to the token-light `itemVersionSummary`. A RouteMapper can't transform responses, so the dispatcher applies the same projection; `full: true` opts into complete rows, mirroring stdio's `--full`. Both modes share one request path, so 404s classify as the kind-aware `item_not_found` (with the ref echoed) either way.
- **Parity, structurally enforced** — the special-case dispatch switch is now an introspectable `specialRoutes()` map (`sync.Once`-cached), and a new parity test drives *every* catalog action through its real ActionFn (link/unlink once per `link_type`), captures the dispatched cmdPath, and fails on any action missing from `routeTable ∪ specialRoutes ∪ itemLinkSpecs` — or any action the fixture can no longer exercise, so coverage can't silently shrink back to a hand-list. Mutation-tested in both directions (removed route → fails naming it; removed fixture key → fails naming the actions). Scope: membership, not behavior — per-route behavior stays with the `TestRoute_*` / real-server tests, which the three new routes bring along.
- **Docs** — CLAUDE.md catalog-version drift fixed (said v0.19; 0.20 shipped in BUG-2302/2305), `version.go` gains a Post-0.20 no-bump note, `instructions.md` mentions history's `full: true`.

## No ToolSurfaceVersion bump

No tool names, action enums, or parameter shapes changed — advertised actions now behave as already documented. A defect fix inside the existing contract, not a contract change; reasoning recorded in `version.go`.

## Review

Codex rounds: R1 two P2s (per-dispatch map rebuild → `sync.Once`; history 404 → kind-aware classifier), one declined (backlinks CLI-hint parity — matches the established behavior of the whole routeSpec item family incl. `item show`; family-wide hint enrichment is a separate concern). R2 confirm caught `full=true` 404s still on the legacy classifier → single request path. R3 fresh-angle → doc fixes, spun off **BUG-2608** (history unbounded on every surface — pre-existing; bounding it needs a catalog shape change), declined encoded-slash identifiers (Pad slug/ref charsets don't admit `/`; family behavior, not this diff). R4 confirm CLEAN.

## Gates

`go test ./...` exit 0; `make lint` 0 issues; mcp package green throughout; `make test-pg` running (result on BUG-2304's trail); real-server smokes drive all three routes end-to-end.


xarmian:
Lead review (artifact + diff): GO. Verified scope against the package: two plain routeSpecs (backlinks carries both paging params; report's omitted filters correctly stay off the URL so endpoint defaults apply), dispatchItemHistory's summary projection with the full=true escape and kind-aware item_not_found in BOTH modes, and the switch→sync.Once-cached map. RULING CONFIRMED on no ToolSurfaceVersion bump: no tool names, action enums, or param shapes changed — advertised actions now behave as already documented, which is a defect fix inside the existing contract; the version.go note is exactly where that reasoning should live. The parity test's unexercised-branch (fixture gaps fail rather than silently shrinking coverage) is the piece that makes this structural instead of another hand-list — good. Checks 7/7 green verified independently. This comment is the on-record approval: flip ready and squash-merge under your grant.

PR #1127, opened 2026-08-17
fix(web): heal collection renames missed by SSE — sync-pass route reconcile + localIndex retag (BUG-2601)

xarmian:
## Root cause

A collection rename changes the slug without touching items, so nothing item-shaped ever re-announces it. Three stranding layers fell out of that one fact:

1. **List route:** delta-sync catch-up covers item changes only, and `/changes` says nothing about renames — a rename-only gap even reports `caught_up`. A client that missed the `collection_updated` SSE kept the dead route slug; slug-keyed fetches 404'd until a manual navigation.
2. **Data (discovered by this fix's own e2e — present on the LIVE SSE path too):** cached `localIndex` rows keep the old `collection_slug` (rows only re-stamp when the item itself changes), so *every* rename-healed route — the existing BUG-2272 SSE heal included — rendered an **empty board** while the sidebar still counted the items.
3. **Full-page item route** (the bug body's own example, `/user/ws/OLDSLUG/item-slug`): same missed-SSE strand, no reconciliation.

## Change

- **Pure decision** `resolveSyncRenameTarget` (renameNav.ts, 9 unit tests): heal iff the route slug is DEAD in the live collections list and the loaded snapshot's stable id maps to a new slug; skips foreign mid-navigation snapshots, reused slugs, deleted ids, in-flight rename gotos, empty lists.
- **List route:** `reconcileRouteCollectionSlug` on every sync pass (`caught_up` included — that silence IS the bug), sharing the `renameNav` tracker; `replaceState` goto; `navigating.to` guard so a heal can never abort an in-flight user navigation; goto-rejection resets the tracker.
- **Item route:** `reconcileCollectionSegment` in ItemDetail (full-page only), mirroring its SSE branch — `renameOverride` bridge, `replaceState`+`noScroll` goto preserving item slug + query, pre-await `collectionGen` fence, `destroyed` guard, `$state`-proxy-identity cleanup token.
- **Rows:** `localIndex.retagCollection(ws, collectionId, newSlug, userId)` re-stamps rows by stable collection id (search + IDB write-through, `upsert` pattern), called from a new workspace-layout global `collection_updated` subscriber (live SSE, any route) and from both sync heals (missed SSE). Renames arriving before hydration are recorded as `pendingRetags` and applied after the warm IDB hydrate — stamped with the recording user's identity so a cold state surviving logout can't steer the next user's cache (null stamp = pre-auth-resolution, adopted by the session's first resolved identity); authoritative server snapshots clear them unapplied.

## Verification

- Three e2e specs, each **verified failing on a control build** of the unfixed sources: missed-SSE list route (aborted `/events` + strand asserted *inside* the hidden window, so a heal firing on the wrong signal fails), missed-SSE item route, live-SSE retag (route heals AND board renders — the leg that exposed layer 2, screenshot ground truth: sidebar `Chores (1)`, board "No chores yet").
- 17 renameNav unit tests; full vitest 1638 passed; svelte-check 0 errors; 6/6 e2e across desktop+mobile projects. Web-only diff.

## Review

8 codex rounds. Notables: the tautological post-await fence, the `$state` raw-vs-proxy identity trap, the navigating-abort hazard. Declined with recorded reasoning: retag O(rows) scan (rare human event, `removeByCollection` precedent), duplicate `loadCollections` (layout-wide pattern, seq-guarded store), IDB fire-and-forget persist race (pre-existing pattern shared with `upsert()` — filed as **BUG-2609** with a fix sketch).

### DIFF
--- a/web/e2e/bug-2601-sse-rename-retag.spec.ts
+++ b/web/e2e/bug-2601-sse-rename-retag.spec.ts
@@ -1,0 +1,48 @@
+import { expect } from '@playwright/test';
+import { test } from './fixtures';
+
+/**
+ * BUG-2601 (retag half) — the LIVE SSE rename path. BUG-2272 re-targets
+ * the route, but cached localIndex rows keep the OLD `collection_slug`
+ * (rows are only re-stamped when the item itself changes, and a rename
+ * touches no items), so the healed route rendered an EMPTY board while
+ * the sidebar still counted the items. Discovered by this spec's
+ * pre-fix run during BUG-2601; fixed by `localIndex.retagCollection`
+ * called from the workspace layout's global `collection_updated`
+ * subscriber. The sibling spec (bug-2601-sync-rename-heal) covers the
+ * missed-SSE path, which retags at the sync-heal site instead.
+ */
+test('BUG-2601: live SSE rename re-targets the route AND the board still renders its items', async ({ page, request, fixture }) => {
+	const auth = { Authorization: `Bearer ${fixture.apiToken}` };
+	const slug = `b2601p-${Date.now().toString(36)}`;
+	const wsResp = await request.post('/api/v1/workspaces', {
+		headers: auth,
+		data: { name: 'BUG-2601 probe', slug, template: 'startup' },
+	});
+	expect(wsResp.ok(), await wsResp.text()).toBeTruthy();
+	const ws = (await wsResp.json()) as { slug: string };
+	try {
+		const itemResp = await request.post(`/api/v1/workspaces/${ws.slug}/collections/tasks/items`, {
+			headers: auth,
+			data: { title: 'probe item' },
+		});
+		expect(itemResp.ok()).toBeTruthy();
+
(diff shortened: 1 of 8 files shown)


xarmian:
Lead review (artifact + diff): GO. Read the pure decision, the localIndex retag machinery, the layout SSE case, and the collection-page heal end-to-end. The two scope growths are the fix being honest — the empty-board discovery means the pre-existing BUG-2272 SSE heal was also shipping a broken landing, and the item-route gap was the filed bug's own example. Particulars that earn the go: resolveSyncRenameTarget's reused-slug skip (leaving ambiguity to the SSE path instead of yanking a live route), pendingRetags' clear-unapplied-on-authoritative-snapshot rule with the cross-user gate and the null-stamp same-session carve-out, the navigating.to yield, and the goto-failure renameNav reset so a poisoned tracker can't veto future heals. Control-build-verified failing e2e for all three specs is exactly the right evidence shape. Checks 7/7 green verified independently. On-record approval: flip ready and squash-merge under your grant.

PR #1128, opened 2026-08-17
fix(web): route ItemDetail collection writes through a semantic adopt gate (BUG-2602)

xarmian:
## Root cause

Seven sites assign ItemDetail's `collection` snapshot under fences that order *starts* (`collectionGen`), and loadData's cross-collection escape hatch (`collGen === collectionGen || collection?.id !== fetched.id`) deliberately admitted any generation-stale write that fetched a different collection. A loadData continuation spanning a cross-collection **move** could therefore restore the SOURCE collection over the freshly adopted TARGET — the live item (itemGen-fenced) kept the move, so the pane rendered the item against the wrong collection's schema. Filed from BUG-2178's codex review; the shape predates it.

## Change

- **`shouldAdoptCollection`** (pure, 9 unit tests, mutation-resistant per review): a snapshot disagreeing with the **live item's `collection_id`** is vetoed regardless of freshness — id, not slug, so renames still apply and a reused slug can't satisfy it (this also closes a latent foreign-write in the SSE `collection_updated` refresh, which fetches by slug); same-collection refreshes keep newest-started-wins; the legitimate cross-collection correction the old hatch existed for still lands when the live item agrees.
- **`adoptCollection`** routes all seven write sites (loadData ×2, SSE refresh, refreshCollectionIfMoved, the BUG-2601 rename heal, quick-action + edit-modal callbacks). Empty-string `collection_id` normalizes to no-anchor.
- **`adoptOrConvergeToLiveCollection`**: on the veto path an embedded pane whose `collection` was still null (fresh mount — `refreshCollectionIfMoved`'s `!collection` guard skips there too, found by runtime instrumentation) converges on the live item's collection instead of being left schema-less; both loadData branches surface a schema-less outcome as the load error and re-check `myGen` after the new await before any singleton writes; the singleton claim (`setActiveItem`/editor reset) now also requires `myItemGen === itemGen`.

## Verification

e2e reproduces the filed race deterministically (route-hold on the realColl fetch, API move mid-hold with a hard `collection_id === tgt.id` oracle on the pane's refetch, release): the **control build renders the moved item against `SrcMarkerField`'s schema** — the filed bug verbatim, screenshot-verified; the fixed build converges on `TgtMarkerField`. Full vitest 1647; svelte-check 0 errors.

## Review

5 codex rounds (R2/R4 confirm CLEAN, R5 fresh-angle only doc nits + a stale-local-main scope false-positive). Substantive catches folded in: post-converge `myGen` re-checks before global writes, schema-less error surfacing on both branches, `convergeGen` captured pre-fetch, itemGen gate on singleton claims, empty-string anchor normalization, hard e2e oracle.

### DIFF
--- a/web/e2e/bug-2602-collection-adopt-race.spec.ts
+++ b/web/e2e/bug-2602-collection-adopt-race.spec.ts
@@ -1,0 +1,170 @@
+import { expect, type APIRequestContext } from '@playwright/test';
+import { test, type SuiteFixture } from './fixtures';
+
+/**
+ * BUG-2602 — ItemDetail's collection writes raced under fences that
+ * ordered STARTS, and loadData's cross-collection escape hatch admitted
+ * any generation-stale write that fetched a different collection. A
+ * loadData continuation spanning a cross-collection MOVE could therefore
+ * restore the SOURCE collection over the freshly adopted TARGET — the
+ * live item (itemGen-fenced) kept the move, so the pane rendered the
+ * item against the WRONG collection's schema.
+ *
+ * Deterministic construction (route-hold, same technique as
+ * pane-collection-migration-race.spec.ts):
+ *
+ *   1. An embedded pane is opened via a cross-collection `?item=` — the
+ *      route is the suite workspace's `tasks`, the item lives in SRC —
+ *      which sends loadData down its realColl branch: `GET
+ *      /collections/<src>`. That GET is HELD by route interception.
+ *   2. While held, the item is MOVED src → tgt via the API. The pane's
+ *      live SSE delivers item_updated; refreshCollectionIfMoved fetches
+ *      TGT (not held) and adopts it — the pane renders TGT's schema
+ *      (its marker field label).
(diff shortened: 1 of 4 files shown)


xarmian:
Lead review (artifact + diff): GO. Read shouldAdoptCollection's decision table and the adoptCollection/adoptOrConvergeToLiveCollection pair end-to-end. The semantic-veto-by-stable-id is the right invariant (the pane renders the LIVE item's collection), and the bonus closure is real: a reused-slug SSE refresh returning a foreign collection now vetoes on id disagreement — including on the converge path's own by-slug fetch, which is what makes the fallback safe to attempt at all. Pre-fetch convergeGen capture + post-await loadGeneration re-check close the orderings; the ||-null empty-string anchor normalization is correctly reasoned. Control-build e2e rendering the filed bug verbatim + mutation-resistance pass + CI 7/7 (verified independently) meets the bar. On-record approval: flip ready and squash-merge under your grant.

PR #1129, opened 2026-08-17
fix(store): orphan GC claim protocol — writer reference stamps + conditional row-first deletes (BUG-2415)

xarmian:
## Root cause

`runOrphanGCSweep` scanned content for `pad-attachment:` references, then deleted the **blob**, then the row — with nothing serializing it against content writers. A reference committing between scan and reclaim left a dangling id or, worse, a surviving row whose bytes were already gone.

## The claim protocol (lead-approved A+B)

- **B — writer-side stamp.** New `attachments.last_referenced_at` (dual-dialect migration 079/057, TEXT RFC3339). `stampAttachmentRefsTx` parses `pad-attachment:` refs (chunked at 400 ids) and stamps the rows **and their variants** (`parent_id IN`) inside the writer's own transaction, executed **before** the content statement — on Postgres the stamp's row locks make a concurrent claim DELETE block until the writer commits, then re-evaluate against the fresh stamp.
- **A — claim-by-delete, row before bytes.** The sweep's row deletion is a conditional DELETE re-asserting reclaimable state in the statement itself: `ClaimNeverAttachedAttachment` (unattached + live + no fresh own-stamp + no fresh **parent** stamp) and `ClaimSoftDeletedAttachment` (still deleted + still past grace — a mid-sweep restore now survives too). The blob is reclaimed **only** after a successful claim, so a surviving row implies surviving bytes; a `BlobDeleteFailures` counter surfaces the inverted failure mode.
- **`orphanGCRefStaleWindow` (15 m)** is documented as a correctness parameter: the LIKE scan still guards long-lived references, so the stamp only covers refs landing after the scan — the window bounds scan-to-claim latency plus a maximally stalled writer transaction, and is explicitly the bound on the >window stamp-to-commit residual.

## Coverage enumeration (constraint 2 — sites, not prose)

Every persistence path that can write a `pad-attachment:` ref, mapped to its stamp call:

| Path | Funnel | Stamp site |
|---|---|---|
| Item create (web/CLI/MCP), artifact import (`createItemChecked`), template seeds | `store.CreateItem` | pre-INSERT, items.go |
| Item PATCH, collab-snapshot flush (`?source=collab-snapshot`), version restore, bulk update | `UpdateItemWithPreCheck/WithParentLink` core | pre-UPDATE, items.go (fields_patch already resolved) |
| Move with `--field` overrides | `MoveItemWithPreCheck` | pre-UPDATE, items.go |
| Comment create / edit | `CreateComment` / `UpdateComment` (both now transactional) | pre-INSERT/UPDATE, comments.go |
| Workspace import | `ImportWorkspace` item + comment loops | pre-INSERT ×2, export.go |
| Cross-workspace copy | clones are inserted **attached** (`item_id` set) → not orphan-eligible | n/a by construction |
| Documents table (legacy) | invisible to scan AND stamp — **pre-existing**, filed **BUG-2614** | out of scope |
| Bundle remap skips comment bodies | **pre-existing**, filed **BUG-2615** | out of scope |

## Accepted residuals (documented at `stampAttachmentRefsTx`)

1. Claim **commits first** → the writer's stamp matches zero rows and the committed text carries a dangling ref — identical to referencing any already-deleted attachment; deliberately not a save failure (zero-rows also describes legitimate unknown/foreign/typo ids).
2. A writer transaction whose stamp-to-commit span exceeds the window — which is why the window is sized to a pathologically stalled transaction.

## Verification

- Sweep-level: fresh-stamp row survives (row **and** blob) with an aged-stamp counterfactual arm — **verified discriminating against a compiling control build of the old sweep order** (control reclaims the fresh-stamped row). Variant test with a both-guards-disabled total-loss control (single-guard mutations pass by design — belt and suspenders proven separately).
- Store-level: every claim predicate leg with counterfactuals, the filed-race sequence, stamp wiring on all sites incl. workspace scoping and variant stamping.
- `go test ./...`, `make lint`, `make test-pg` green at every round; 5 codex rounds (R2/R5 confirm CLEAN; substantive catches folded: coverage gaps → move+import stamps, variant hole → parent-aware scan + claim + variant stamping, stamp ordering → pre-content, chunking, observability counter).

## For PLAN-2411 / PLAN-2416

A true claim is irrevocable (noted on both Claim methods): recovery surfaces can restore only rows the claim hasn't taken — restores that clear `deleted_at` before the claim commits now survive the sweep by predicate.

### DIFF
--- a/internal/server/orphan_gc.go
+++ b/internal/server/orphan_gc.go
@@ -15,6 +15,27 @@ import (
 const (
 	defaultOrphanGCInterval = 24 * time.Hour
 	defaultOrphanGCGrace    = 30 * 24 * time.Hour
+
+	// orphanGCRefStaleWindow is a CORRECTNESS parameter, not a tuning
+	// knob (BUG-2415). A never-attached row is only claimable when its
+	// last_referenced_at stamp is older than this window.
+	//
+	// What it must cover: the stamp is NOT a lease on long-lived
+	// references — those are caught by the content LIKE scan
+	// (AttachmentReferenced) that runs immediately before the claim.
+	// The stamp only has to cover references that commit AFTER that
+	// scan, i.e. within the scan→claim gap of a single sweep iteration
+	// (milliseconds) PLUS the full duration of the writer transaction
+	// that carries the stamp — including a pathologically stalled one
+	// (lock waits, an operator-paused Postgres, a laptop suspending
+	// mid-commit). 15 minutes exceeds any plausible single write
+	// transaction by orders of magnitude while delaying reclamation of
+	// a genuinely-orphaned row by at most one extra sweep against a
+	// 30-day grace period — the asymmetry is entirely in favor of
+	// generosity. If you are tempted to lower this for faster tests,
+	// inject the cutoff in the test instead; the constant guards
+	// production correctness.
+	orphanGCRefStaleWindow = 15 * time.Minute
 )
 
 // orphanGCResult records what one sweep accomplished. Returned from
@@ -77,7 +98,8 @@ func (s *Server) runOrphanGCSweep(ctx context.Context, graceCutoff time.Time) (*
 		// + comment bodies before reclaiming so the GC doesn't destroy
 		// a legitimate reference. Codex P1 on PR #307 round 1;
 		// comment-body coverage added for IDEA-1650.
-		if a.ItemID == nil && a.DeletedAt == nil {
+		neverAttached := a.ItemID == nil && a.DeletedAt == nil
+		if neverAttached {
 			referenced, err := s.store.AttachmentReferenced(a.WorkspaceID, a.ID)
 			if err != nil {
 				slog.Warn("orphan GC: ref-scan failed",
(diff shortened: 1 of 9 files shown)


xarmian:
Lead review (artifact + diff): GO. Read the claim protocol, the sweep rewrite, and the stamp chokepoints end-to-end. The construction is right: conditional-DELETE-as-claim serializes against in-tx writer stamps at the database, row-before-bytes makes a surviving row imply surviving bytes, and the two documented residuals are genuinely irreducible (claim-committed-first = dangling ref, same as any deleted-id reference; stamp-to-commit exceeding the window = exactly what the 15m sizing criterion bounds — and the constant's comment saying 'inject the cutoff in tests, don't lower this' is precisely how a correctness parameter should defend itself). The parent-aware variant handling closing a pre-existing referenced-original-loses-its-thumbnails hole alongside, with the parent_id IN (...) lock making claimed thumbnails wait on their own row — good catch chain across the codex rounds. Both review constraints carried as asked (window rationale at the constant; coverage table in the body with the two pre-existing gaps filed as BUG-2614/2615 rather than laundered in). Control-build-discriminating tests at sweep, variant, and predicate level; CI 7/7 verified independently. On-record approval: flip ready and squash-merge under your grant.

PR #1130, opened 2026-08-17
feat(server): armed-session declaration + push delivery filter (PLAN-2613 S1)

xarmian:


xarmian:
On-record approval (same-account form): implementer plan approved on TASK-2616 trail before build; diff reviewed by lead (gate placement, both delivery paths, count semantics); codex R1 multi-angle CLEAN + R2 adversarial bypass hunt clean on the gate with two pre-existing documented observations declined; CI 7/7. Merging under the PLAN-2613 build.

PR #1131, opened 2026-08-17
fix(web): source-identity guards on the SSE non-sync listeners (BUG-2611)

xarmian:
## Root cause

`close()` does not retract already-queued event tasks, so on a fast workspace switch a torn-down `EventSource`'s queued events can fire after the next workspace's source exists. BUG-2540 guarded `onopen`/`onerror`/`connected` with a `source === eventSource` identity check; the remaining listeners had none. Consequences ranked (from the staging re-verify on the item): a stale `unauthorized` **closed the new workspace's live EventSource**, flipped the status indicator, and cleared `currentWorkspace` over the old workspace's auth state — nothing reconnected until a navigation; stale `sync_required`/`items_bulk_updated` triggered spurious reconciles fanned onto the new workspace's BroadcastChannel; stale item events leaked cross-workspace into the new workspace's callbacks (including a mislinked creation toast).

## Change

The same one-line guard on all four listeners (`sync_required`, `items_bulk_updated`, `unauthorized`, the `ITEM_EVENTS` loop). `unauthorized` additionally closes its **own** source rather than whatever `eventSource` currently points at — the guard has just proven they're the same object, and the old shape is exactly what made the stale path destructive.

## Verification

Unit harness on BUG-2540's stubbed-EventSource pattern (`sseStaleSourceGuards.svelte.test.ts`): drive a fast A→B switch, fire each event type on the torn-down A source — zero dispatch into B — with a **live-source control arm per leg** so a guard that silences both worlds cannot pass. All four legs verified failing on the unguarded build (4/4 red on control, 4/4 green fixed). Full vitest 1651 passed; svelte-check 0 errors. Codex round 1 CLEAN on a multi-angle probe (reconnect churn, no-locks fallback, the unauthorized rewrite's equivalence, harness vacuity).

### DIFF
--- a/web/src/lib/services/sse.svelte.ts
+++ b/web/src/lib/services/sse.svelte.ts
@@ -249,6 +249,13 @@ function createSSEService() {
 
 		// Handle sync_required: server's replay buffer couldn't cover the gap.
 		source.addEventListener('sync_required', () => {
+			// Source-identity guard (BUG-2611, mirroring BUG-2540's
+			// onopen/onerror/'connected' guards): close() does not retract
+			// already-queued event tasks, so on a fast workspace switch a
+			// torn-down source's event can fire after the next workspace's
+			// source exists — and would dispatch into (and broadcast onto)
+			// the NEW workspace's channel.
+			if (source !== eventSource) return;
 			dispatchSyncRequired();
 			broadcast({ type: 'sync_required' });
 		});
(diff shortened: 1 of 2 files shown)


xarmian:
Lead review (artifact + diff): GO. The unauthorized rewrite is the load-bearing piece — close-own-source after the guard proves it's live, versus the old close-whatever-eventSource-points-at that let workspace A's stale auth event tear down B. Guards match the 2540 idiom; per-leg control arms close the silenced-both-worlds hole. Single codex round is proportionate to the diff. On-record approval: flip ready and squash-merge under your grant.

PR #1132, opened 2026-08-17
fix(web): portal the pane action-bar menus — anchored panels clipped against the pane's scroll container (BUG-2610)

xarmian:
## Root cause

In split view `.item-pane` is an `overflow-y:auto` scroll container — which computes `overflow-x:auto` too — and the quick-actions (⚡) and ⋯ menus were **anchored** panels inside it. A right-aligned panel opening from the pane's action bar extends left past the pane's edge, so the container paint-clipped it mid-text (Dave's screenshots: "age actions" for "Manage actions", truncated tagline).

## Change

- Both menus flip to the Menu component's **portal mode** — built precisely to escape overflow containment (fixed coords portaled to `<body>`, viewport clamping, flip-when-cramped), and already the mode of every board-card menu. Widths mirror content (QA 288 / ⋯ 260).
- **Portal scroll-dismiss scoped to anchor-moving containers** — the any-scroll dismissal closed open menus whenever *unrelated* containers scrolled (live SSE re-layout churn). Found because the capstone move e2e went parallel-unstable on this branch while a control build ran clean — treated as signal, not flake. Dismiss now fires only when the scrolled container contains the trigger; panel-internal and `exempt` surfaces (nested emoji picker) are exempt.
- **Portaled-menu key leak closed** (codex): the collection route's nav-key bail checks `target.closest('.item-pane')`, which a portaled row no longer matches — j/k/Enter drove list navigation under an open menu. `[role="menu"]` joins the bail (also closing the same **pre-existing** leak for board-card portal menus), plus `stopPropagation` on Menu-handled arrow keys.

## Verification

- Regression e2e with a **paint-level oracle**: clipping doesn't shrink `getBoundingClientRect`, so `elementFromPoint` just inside the panel's left edge must resolve to the panel; per-menu geometry preconditions keep it non-vacuous. Verified failing on the anchored control build with the exact reported symptom.
- Blast radius swept: 14 e2e menu-row lookups page-scoped (portaled panels live in `<body>`; the scoped trigger click ties each menu to its column), QA unit-test queries re-scoped to document. Targeted pane suites 34/34 ×3 under parallel load; vitest 1651; svelte-check 0.
- **Full-suite flake transparency**: five full-suite runs (3 branch / 1 control / 1 branch) each flaked 1-2 tests from the documented parallel-load family (content-link popups, capstone move — the latter also on **control**, a11y j/k, viewer-owners), always passing isolated and in groups; tally + triage method recorded on BUG-2447. The one genuinely diff-caused parallel failure found this session was the scroll-dismiss regression above — fixed, not excused.

## Review

Codex R1: key-leak (fixed route-side), exempt-scroll gap (fixed), per-menu e2e preconditions (fixed); R2 confirm caught two remaining pane-scoped graph lookups (fixed). Shadow-DOM `contains()` awareness declined — no shadow DOM in the app.

### DIFF
--- a/web/e2e/bug-2610-pane-menu-clip.spec.ts
+++ b/web/e2e/bug-2610-pane-menu-clip.spec.ts
@@ -1,0 +1,82 @@
+import { expect, type Page } from '@playwright/test';
+import { test, type SuiteFixture } from './fixtures';
+
+/**
+ * BUG-2610 — in split view the pane's quick-actions (⚡) and ⋯ menus were
+ * ANCHORED panels inside `.item-pane`, an `overflow-y:auto` scroll
+ * container (which computes `overflow-x:auto` too). A right-aligned panel
+ * opening from the pane's action bar extends LEFT past the pane's left
+ * edge, so the container clipped it mid-text (Dave's screenshots on the
+ * item: "age actions" for "Manage actions"). Both menus now use the Menu
+ * component's portal mode, which exists precisely to escape overflow
+ * containment.
+ *
+ * The oracle is a PAINT-level probe, because clipping doesn't shrink
+ * getBoundingClientRect: `document.elementFromPoint` just inside the
+ * panel's LEFT edge must resolve to the panel itself. On the anchored
+ * control build the panel's left region lies outside the pane's clip, so
+ * the probe hits whatever is underneath (the board) — which is exactly
+ * what a user's click there did.
+ */
+
+function authHeaders(fixture: SuiteFixture) {
+	return { Authorization: `Bearer ${fixture.apiToken}` };
+}
+
+/** Probe: does the point just inside the panel's left edge actually hit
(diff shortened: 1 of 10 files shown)


xarmian:
Lead review (artifact + diff): GO. The contains-the-trigger dismiss predicate is the correct physical model (a container that can't move the anchor can't stale fixed coords), the panel-internal and exempt-surface carve-outs are properly reasoned, and the [role=menu] route-bail + handled-key stopPropagation close the leak for board-card portal menus too — a real pre-existing fix riding along honestly. Paint-level elementFromPoint e2e with geometry preconditions verified failing on the anchored control is the right instrument for a clipping bug. Flake transparency with control-build triage recorded on BUG-2447 is the standard we should keep. CI 7/7 incl. full E2E verified. On-record approval: flip ready and squash-merge under your grant.

PR #1133, opened 2026-08-17
fix(server): refuse a PATCH carrying both a fields hierarchy key and top-level parent_id (BUG-2594)

xarmian:
## Root cause

`{"fields_patch": {"parent": ""}, "parent_id": "<id>"}` in one PATCH applied **both** hierarchy writes: `extractParentLink` staged the `item_links` write (including the empty-string clear) while `ItemUpdate.ParentID` stamped the `parent_id` column unconditionally in the same transaction — link cleared, column re-parented, `unparentedItemPredicate` still saw a parent. Silent inconsistent state, no error. Pre-existing since the `parent_id` column landed on `ItemUpdate` (2026-07-08); surfaced by BUG-2078's review.

## Change

Both update paths (full `fields` + `fields_patch`) refuse the pair with a `validation_error` naming both keys — **refused, not silently resolved**, per the clear_parent contract family's standing rule (v0.19). The audit found the shape is raw-HTTP-only: no first-party client sends top-level `parent_id` on item update (CLI resolves `--parent` into the patch; web client and MCP catalog never carry it) — confirmed independently by review across bulk/restore/move/import/MCP dispatchers. Solo `parent_id` and solo fields-hierarchy writes are deliberately unchanged (BUG-2379 tracks the adjacent undeclared-override family); the schema-declared `parent`-field case stays an ordinary field write per the v0.19 shadowing rule.

## Verification

Six handler tests: refusal on clear+id, set+id, the `plan` alias, and the full-`fields` sibling path — **each verified failing (200) on the unguarded control build** — plus both solo-write controls, envelope-code and both-keys-named assertions. Full `internal/server` suite green. Codex round 1: clean on caller audit, guard placement, and the schema-shadowed precedence; its test-tightening note applied.

### DIFF
--- a/internal/server/handlers_items.go
+++ b/internal/server/handlers_items.go
@@ -956,6 +956,19 @@ func (s *Server) handleUpdateItem(w http.ResponseWriter, r *http.Request) {
 		if !extractOK {
 			return
 		}
+		if parentProvided && input.ParentID != nil {
+			// BUG-2594: a hierarchy write in fields ("parent"/"plan" — the
+			// canonical item_links surface) and a top-level "parent_id"
+			// (a raw column stamp no first-party client sends on update)
+			// in one request are two competing hierarchy writes. The old
+			// behavior applied BOTH — clearing the link while stamping
+			// the column, leaving silently inconsistent state. Refused,
+			// not silently resolved, per the clear_parent contract
+			// family's standing rule (v0.19).
+			writeError(w, http.StatusBadRequest, "validation_error",
+				`Request carries both a "parent"/"plan" key in fields and a top-level "parent_id" — two competing hierarchy writes. Send exactly one.`)
+			return
+		}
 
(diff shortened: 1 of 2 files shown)


xarmian:
Lead review (artifact + diff): GO. Additive-only guard on both update paths, refused-not-resolved per the v0.19 rule with both competing keys named in the envelope, and the scope discipline is right — solo writes untouched with BUG-2379 owning the adjacent family, schema-declared parent staying an ordinary field per the shadowing rule. Four control-verified refusal legs + two solo-write controls is proportionate coverage for a guard this size. The instrument-check process note (escaped-vs-raw substring before concluding the guard didn't fire) is PATTE-135 discipline applied to your own test harness — worth the minute it cost. CI 7/7 verified. On-record approval: flip ready and squash-merge under your grant.

PR #1134, opened 2026-08-17
fix(store): attachment delete vs thumbnail derivation race — atomic cascade, locked conditional insert, orphaned-variant GC class (BUG-2388)

xarmian:
## Root cause

Deleting an attachment mid-derivation could mint a live, unreachable variant row under a tombstoned parent: the delete cascade tombstoned original and variants in **separate statements**, and derivation checked parent liveness once, then inserted unconditionally. The leaked row was UI-invisible, counted toward quota forever, and **no GC class could reclaim it** — the old code's comment claimed a "deleted-parent path" existed; it did not.

## Change (the BUG-2415 claim-by-statement family)

- **Atomic cascade** — `SoftDeleteAttachment` tombstones original + variants in one transaction; a variant-tombstone failure now rolls back rather than leaving a partial family.
- **`CreateAttachmentVariantIfParentLive`** — transaction + dialect-gated `FOR NO KEY UPDATE` parent lock + insert, the `CreateAttachmentForLiveItem` precedent (review round 1 showed a plain `WHERE EXISTS` still races on Postgres — snapshot reads don't wait for a committing cascade; the lock makes whichever transaction commits first observable to the other). `persistThumbnail` cleans up the just-Put blob on refusal with the count, self-registration check, and `Delete` all held under `inFlightHashesMu` (round 2: a complete upload lifecycle could otherwise stale an outside-the-fence count), honoring the operator-configured GC grace.
- **Orphaned-variant GC class** — live variant whose parent is tombstoned/gone; candidate branch + `ClaimOrphanedVariantAttachment`, tried first for live parented rows (an `item_id`-NULL leak would otherwise hide behind a content reference to its dead parent in the never-attached scan). The claim re-reads the variant and **locks the parent** in its own transaction, so a concurrent restore refuses it — a restored original keeps its thumbnails. Retro-reclaims rows already leaked. Lock ordering between the insert tx and the cascade tx reviewed: both take parent before children — no deadlock surface.

## Verification

- The filed race pinned **deterministically**: `persistThumbnail` called with a pre-delete parent snapshot (exactly what the async goroutine holds) — the control build mints the leaked row with the test's own message; the fixed build refuses, mints nothing, and the blob-cleanup assertion (deterministic hash recompute + `Stat`) proves the bytes went too.
- Retro-reclaim sweep test with a deliberately **attached** leak fixture so only the new class can reclaim it (control build: "leaked live variant survived the sweep"), plus a restore-wins leg.
- Store-level restore-refusal-**at-claim-time** test with tombstoned-counterfactual and hard-gone-parent legs (round 2 observed the e2e restore leg only proved candidate non-selection).
- `go test ./...` 0 / `make lint` 0 / `make test-pg` 0 at both round tips. Codex rounds: R1 four findings fixed, R2 one, R3 confirm CLEAN.

### DIFF
--- a/internal/server/handlers_attachments_thumbnails.go
+++ b/internal/server/handlers_attachments_thumbnails.go
@@ -275,16 +275,31 @@ func (s *Server) persistThumbnail(
 		// The parent was deleted between the derivation's early liveness
 		// check and this insert (BUG-2388) — the conditional insert is
 		// the fix for the race that used to mint a live variant row
-		// under a tombstoned parent. Clean up the just-Put blob while we
-		// still hold the in-flight hash fence, unless another live/
-		// in-grace row shares the hash (same dedupe protection as the
-		// GC sweep). A cleanup failure only strands bytes — never a row.
-		others, cErr := s.store.CountProtectingAttachmentsForHash(hash, "", time.Now().Add(-defaultOrphanGCGrace))
-		if cErr == nil && others == 0 {
-			if dErr := store.Delete(ctx, storageKey); dErr != nil {
-				slog.Warn("thumbnails: orphan blob cleanup failed",
-					"storage_key", storageKey, "error", dErr)
+		// under a tombstoned parent. Clean up the just-Put blob, unless
+		// another live/in-grace row shares the hash (same dedupe
+		// protection as the GC sweep, using the CONFIGURED grace — a
+		// longer operator grace must keep a still-restorable peer's
+		// bytes). The count runs first; the in-flight re-check + Delete
+		// then hold inFlightHashesMu, mirroring the sweep's fence, so a
+		// concurrent upload registering this hash between our check and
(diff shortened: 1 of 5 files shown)


xarmian:
Lead review (artifact + diff): GO. CreateAttachmentVariantIfParentLive follows the CreateAttachmentForLiveItem precedent faithfully — tx + dialect-gated FOR NO KEY UPDATE with the gate documented as a correctness requirement, refuse-on-tombstoned with no row minted, and the lock-ordering review (parent before children in both the cascade and the claim) is the analysis that needed to exist for a deadlock-free pair. The retro-reclaim GC class with restore-refusal-at-claim-time closes the historical leaks, not just the go-forward race, and the deliberately-attached fixture making only the new class discriminate is careful test design. Three control-verified levels again; CI 7/7 verified. On-record approval: flip ready and squash-merge under your grant.

PR #1135, opened 2026-08-17
fix(web): share pages render attachment refs as honest placeholders (BUG-2389)

xarmian:
## Problem

The public share route (`/s/{token}`) rendered item content with a bare `marked()` call, so `pad-attachment:` references fell through as broken `` tags and dead links. The other half of BUG-2389 named `CommentThread.svelte` — which turned out to be unmounted dead code.

## Fix

1. **`CommentThread.svelte` deleted** — grep proved nothing imports it (its only references were prose comments, now cleaned in `ItemDetail.svelte` and `handlers_artifact_import.go`), so its half of the bug resolves by deletion.
2. **Share route renders attachment-aware.** New opt-in wrapper `renderMarkedWithAttachments()` threads an attachment context (resolver + missing-hook) into the existing marked renderer hooks, without `renderMarkdown`'s wiki-link machinery (wiki-links stay deliberately inert on share pages). The share page calls it with a null resolver and the new `renderAttachmentUnavailable()` hook, so every ref renders an honest chip — *"not available on shared pages"* as **visible text** (share pages get read on touch devices; a `title` tooltip alone doesn't exist there), deliberately NOT the false "missing or has been deleted" wording. Chip CSS is global in `app.css` so it reaches both `.item-content` and collection-share expansions.
3. **Sanitization posture unchanged**: the wrapper returns unsanitized HTML; the share page keeps its single DOMPurify pass. The Go guard test (`TestSharePageDoesNotImportCollab`) is updated per its own instructions to pin the new safe path.
4. **Module context hardened** (codex): both the wrapper and `renderMarkdown` now save/restore the shared render context instead of clearing to null, so nested renders can't strand an outer parse.

**Deliberately out of scope:** the token-scoped attachment byte endpoint that would serve real images on share pages (2b) — that's a new unauthenticated ACL surface, tracked separately pending approval. A real resolver through the same wrapper is the plug-in point (pinned by test).

## Tests

- `markdown.shareAttachments.test.ts` (node): placeholder for image + link refs, no false wording, bare-`marked()` opt-in control, real-resolver full resolution, adversarial alt escaping, re-entrancy (nested wrapper AND nested `renderMarkdown`), context restoration.
- `markdown.shareAttachments.svelte.test.ts` (jsdom): the placeholder survives the share page's default-config DOMPurify pass intact; hostile alt stays neutralized post-sanitize.
- `bug-2389-share-attachment-placeholder.spec.ts` (e2e): real upload → item content ref → item share link → anonymous context visit; asserts 2 placeholder chips, visible note text, title, no `img[src^="pad-attachment:"]`, no false wording. **Verified failing on the pre-fix build** (control leg: pre-fix rendered zero placeholders).

Gates: vitest 92 files / 1663 passed · svelte-check 0 errors · full playwright 191 passed / 0 failed · `go test ./...` clean · lint 0 issues. Codex review converged CLEAN at round 4 (findings fixed along the way: context re-entrancy, adversarial/DOMPurify test legs, stale prose refs, unstyled chip, expansion-scope CSS).

BUG-2389

https://claude.ai/code/session_017jD6t1zjxGSq47SQpZfp1V


xarmian:
Lead review (artifact + diff): GO, contingent only on CI green (still pending at review time). Verified the CommentThread dead-code claim with an independent grep — only prose references on main, no import or mount — so deletion over wire-up is the right disposition and was correctly surfaced as a scope change with evidence rather than silently absorbed. The honest-chip wording (visible text, not title-only; explicitly not the false 'missing or deleted') is CONVE-98-grade honesty pointed at end users. Sanitization posture preserved (wrapper stays unsanitized, single DOMPurify pass at the page, guard test updated per its own instructions) with jsdom legs proving the chip survives the real sanitizer config against hostile alt text — that plus the context save/restore re-entrancy fix were the load-bearing codex catches. The pinned real-resolver plug-in point means the (2b) endpoint lands as a resolver swap when Dave approves it. On-record approval: flip ready and squash-merge under your grant once checks complete.

Issue #1066, opened 2026-08-08
MCP: pad_item create/update silently ignores valid fields.status / fields.priority writes (no error, updated_at still bumps)


xarmian:
## Summary

`pad_item` `create` and `update` calls made through the **MCP server** silently ignore `fields.status` (and sometimes `fields.priority`) writes. No error or warning is returned — the call succeeds, `updated_at` is bumped, and the returned item still carries the old field value. This happens even when the written value is **valid per the collection's declared schema enum** (verified against `pad_collection list` output). The behavior is intermittent across sessions: field writes from some sessions stick, from others they are ignored for the entire session.

## Environment

- Client: Claude Code driving the **pad MCP server** (tool surface v0.14 per the server's instructions block; items created this way show `"source": "cli"`)
- Workspace: `house-of-voi` (`workspace_id: 374f64f3-04dc-40d7-ab45-09892302cf0b`)
- Collections involved: Tasks (`b8fa3da9-e53b-4a56-99c7-...` prefix TASK), Plans (prefix PLAN)
- No `expected_updated_at` passed in any of the calls below (so it isn't an optimistic-concurrency 409 — and no structured `update_conflict` error was returned either)

## Repro from session on 2026-08-08 (UTC timestamps from the tool responses, for log correlation)

Collection schema at the time, from `pad_collection list` — Tasks status enum:

```json
{"key": "status", "type": "select", "default": "backlog", "options": ["backlog", "active", "review", "done"], "required": true, "terminal_options": ["done"]}
```

**1. Create — invalid enum value silently replaced by default (no validation error):**

- `2026-08-08T16:57:58Z` — `pad_item action=create collection=tasks` with `fields: {"priority": "medium", "status": "open"}` (note: `open` is not in the Tasks enum — my mistake, but see behavior)
- Result: item created as **TASK-598** (`id: 1f99edc8-69f2-4d60-8d79-fb33d4482b27`) with `"status": "backlog"`. No error, no warning that the requested value was dropped.

**2. Update — invalid enum value silently ignored:**

- `2026-08-08T17:14:35Z` — `pad_item action=update ref=TASK-598 fields={"status": "doing"}` (`doing` is also not in the enum)
- Result: success response, `updated_at` bumped to `17:14:35Z`, returned item still `"status": "backlog"`. No error.

**3. Update — VALID enum value silently ignored (the core bug):**

[...]

PR #1152, opened 2026-08-18
fix(server): stop counting disabled conventions as completed work

### LINKED ISSUE #1049: standup/report count disabled conventions as completed work — listTerminalItemsSince ignores schema terminal_options
## Summary

Disabling a convention is reported as **completed work**. `pad project standup` lists disabled conventions under `completed`, and `pad project report`'s completed counter includes them.

This is two separable things, one mechanical and one semantic:

1. `listTerminalItemsSince()` iterates `models.DefaultTerminalStatuses` unconditionally and **ignores the collection's schema-declared `terminal_options`**, even though `models.TerminalValuesForDoneField()` exists for exactly this.
2. Even if it honoured the schema, the stock `startup` template declares `terminal_options: ["disabled"]` on Conventions — so *terminal* and *completed work* are the same thing for a collection whose items are rules, not work.

## Steps to reproduce

```bash
pad workspace create "demo" --slug demo --template startup
pad --workspace demo item update CONVE-2 --status disabled
pad --workspace demo project standup --format json | jq '.completed'
```

Actual:

```json
[ { "ref": "CONVE-2", "title": "Never push directly to main", "status": "disabled" } ]
```

Expected: `[]` — no work was completed; a rule was switched off.

`pad project report --format json` counts it too: on a fresh workspace where I disabled 4 seeded conventions and completed nothing, the current day's bucket reads `{"created": 15, "completed": 4}`.

## Mechanism

`internal/server/handlers_project_intel.go`, `listTerminalItemsSince()` (v0.11.0):

```go
for _, status := range models.DefaultTerminalStatuses {
	items, err := s.store.ListItems(workspaceID, models.ItemListParams{
		CollectionIDs: collIDs,
		Fields:        map[string]string{"status": status},
```

`models.DefaultTerminalStatuses` (`internal/models/terminal.go`) is the union list, which includes `"disabled"`, `"archived"` and `"deprecated"`. Its own doc comment scopes it narrowly:

> `DefaultTerminalStatuses` is the fallback list used when a collection's done field has no `terminal_options` declared on its schema.

But the standup/changelog path uses it as the primary source, not a fallback. `TerminalValuesForDoneField(schema, settings)` is never consulted, so a collection that *does* declare `terminal_options` gets no say.

**Verified:** setting `terminal_options: []` on the Conventions `status` field changes nothing — the disabled conventions are still reported as completed. That rules out schema configuration as a fix and localises it to this loop.

Same union list also means an archived **Doc** counts as completed work (`terminal_options: ["archived"]` on the stock Docs collection). I didn't test that one.

## Why it matters

We onboard repos onto Pad **default-deny**: the four `system`-authored conventions the `startup` template seeds are agent-binding config, and a stock `[MUST]` can outrank the repo's own ratified docs. (In our first repo, `CONVE-2 "Never push directly to main"` contradicted three repo documents and sat active unnoticed for 19 days.) So each new workspace disables 3-4 conventions on creation and re-enables only what the repo's rules agree with.

The consequence is that **every repo we add injects 3-4 phantom completions**. In our cross-workspace rollup at 2 workspaces, 4 of the 5 "completed" rows were disabled conventions. It scales linearly with the fleet, and it makes throughput unusable as a signal — the number goes up when you turn rules *off*.

## Suggested fix

For (1): have `listTerminalItemsSince()` resolve terminal values per collection via `TerminalValuesForDoneField()` rather than iterating the global union. Since it already loops per status and is best-effort, this looks like resolving the set per `collID` instead of once globally.

For (2) — the design question, and yours to make the call on: even a schema-correct terminal set conflates "this rule is off" with "this work is done". The cleanest separation is probably a per-collection flag for whether its items represent work (something like `counts_as_work: false` on Conventions/Playbooks/Docs), so standup, changelog and report throughput can skip non-work collections while `ready`/`stale` are unaffected. Happy to send a PR for (1) if you want it split that way.

## Workarounds tried

- **Schema `terminal_options: []` on Conventions** — no effect, per above.
- **Using `draft` instead of `disabled`** to park an unwanted convention — this *does* drop it out of `completed`, but a `draft` convention still appears in the default `pad item list conventions` output, whereas a `disabled` one is hidden. For default-deny that's a bad trade: the visibility is the whole point.
- **Client-side filtering** — what we settled on. Our rollup drops any `standup.completed` entry whose status isn't `done`/`completed`. Note this only works because the standup payload carries per-item `status`; `project report` returns bare per-day counts, so its completed number can't be corrected downstream.

## Environment

- pad `0.9.2` (`7687521`, 2026-07-06), self-hosted daemon, SQLite, single user.
- Source citations above are from tag **v0.11.0** — the code path is unchanged there, but I have not run 0.11.0, so the runtime reproduction is on 0.9.2.

## Minor, possibly by design

`project report`'s completed counter is monotonic within a bucket: moving a task `open → done → open` leaves the count incremented. If the metric is meant to be "transitions into terminal during the window" that's defensible, but combined with the above it means a disabled convention is permanently baked into that day's throughput.

asjdf:
## Summary

Fixes #1049. I checked open PRs on 2026-08-18 (dependabot, #1125 / #1066, #1034 web) — none of them touch `listTerminalItemsSince` or this contract. Upstream `main` still loops `models.DefaultTerminalStatuses` and ignores per-collection `terminal_options`.

Standup, changelog, and `project report` treated a disabled convention as completed work:

1. **Mechanical:** `listTerminalItemsSince` (and the CLI standup/changelog loops) iterated the global `DefaultTerminalStatuses` union instead of `TerminalValuesForDoneField()`. A collection that only declares `shipped` was still scanned for `done`; setting Conventions `terminal_options: []` changed nothing.
2. **Completed-work semantics:** stock Conventions declare `terminal_options: ["disabled"]`. Even after honoring the schema, `disabled` is a non-shipping close, not completed work. Report already had this distinction as `negativeTerminals` (rejected/cancelled/…). I moved that list to `models` (shared with standup/changelog) and added `disabled`. No new `counts_as_work` product flag — that's the issue's part 2, a maintainer call.

What changed:

- `listTerminalItemsSince` resolves positive terminals per collection and keeps the existing CollectionIDs **OR** ItemIDs visibility filter (guest item-grant changelog still works).
- CLI `pad project standup` / `changelog` use the same helper (`listCompletedWorkSince`) so the KEEP IN SYNC contract holds. MCP already proxies the server handlers.
- `GetReport` uses `models.IsNegativeTerminal`, so a disabled convention no longer increments the day's completed count.

## Test plan

- [x] New tests failed on current `main` (disabled CONVE in standup/changelog/report; custom `terminal_options: ["shipped"]` still counted `done`).
- [x] After the fix:
  - `go test ./internal/models/ ./internal/store/ ./internal/server/` — pass
  - Focused cases: `TestProjectStandupEndpoint_DisabledConventionNotCompleted`, `TestProjectStandupEndpoint_HonorsCollectionTerminalOptions`, `TestProjectChangelogEndpoint_DisabledConventionNotCompleted`, `TestGetReport_DisabledConventionNotCompleted`, `TestGetReport_NegativeTerminalNotCompleted`, existing standup/changelog/ACL tests
  - `go vet` on the changed packages
  - CLI package compile-checked (`go build ./cmd/pad` with a stub `web/build`; full `make build` not run — web embed is out of scope)
- [ ] Original CLI repro (`pad workspace create` / `item update CONVE-2 --status disabled` / `project standup --format json`) not run end-to-end here; the HTTP standup/changelog/report tests encode that contract.

### DIFF
--- a/cmd/pad/cmd_project.go
+++ b/cmd/pad/cmd_project.go
@@ -265,6 +265,40 @@ func nextCmd() *cobra.Command {
 	}
 }
 
+// listCompletedWorkSince fetches items that reached a *positive* terminal
+// since cutoff. Terminal values are resolved per collection via
+// models.CollectionCompletedWorkValues — KEEP IN SYNC with
+// server.listTerminalItemsSince (BUG-1049). Best-effort per collection
+// and status: a list error skips that query rather than failing the
+// whole report.
+func listCompletedWorkSince(client *cli.Client, ws string, cutoff time.Time, limit int) []models.Item {
+	colls, err := client.ListCollections(ws)
+	if err != nil {
+		return nil
+	}
+	var out []models.Item
+	for _, c := range colls {
+		field, values := models.CollectionCompletedWorkValues(c.Schema, c.Settings)
+		for _, status := range values {
+			params := url.Values{
+				field:   {status},
+				"sort":  {"updated_at:desc"},
+				"limit": {strconv.Itoa(limit)},
+			}
+			items, err := client.ListCollectionItems(ws, c.Slug, params)
+			if err != nil {
+				continue
+			}
+			for _, item := range items {
+				if item.UpdatedAt.After(cutoff) {
+					out = append(out, item)
+				}
+			}
+		}
+	}
+	return out
+}
+
 // --- standup ---
 
 func standupCmd() *cobra.Command {
@@ -315,26 +349,8 @@ func standupCmd() *cobra.Command {
 				return fmt.Errorf("parsing dashboard: %w", err)
 			}
 
-			// Fetch recently completed items (terminal statuses)
-			doneStatuses := models.DefaultTerminalStatuses
-			var completedItems []models.Item
 			cutoff := time.Now().AddDate(0, 0, -days)
-
-			for _, status := range doneStatuses {
-				items, err := client.ListItems(ws, url.Values{
-					"status": {status},
-					"sort":   {"updated_at:desc"},
-					"limit":  {"20"},
-				})
-				if err != nil {
-					continue
-				}
-				for _, item := range items {
-					if item.UpdatedAt.After(cutoff) {
-						completedItems = append(completedItems, item)
-					}
-				}
-			}
+			completedItems := listCompletedWorkSince(client, ws, cutoff, 20)
 
 			// Fetch in-progress items
 			inProgressItems, err := client.ListItems(ws, url.Values{
@@ -689,25 +705,7 @@ func changelogCmd() *cobra.Command {
 				cutoff = time.Now().AddDate(0, 0, -days)
 			}
 
-			// Fetch completed items across all terminal statuses
-			doneStatuses := models.DefaultTerminalStatuses
-			var allItems []models.Item
-
-			for _, status := range doneStatuses {
-				items, err := client.ListItems(ws, url.Values{
-					"status": {status},
-					"sort":   {"updated_at:desc"},
-					"limit":  {"100"},
-				})
-				if err != nil {
-					continue
-				}
-				for _, item := range items {
-					if item.UpdatedAt.After(cutoff) {
-						allItems = append(allItems, item)
-					}
-				}
-			}
+			allItems := listCompletedWorkSince(client, ws, cutoff, 100)
 
 			// Filter by parent if specified
 			filterParent := parentRef
--- a/internal/models/terminal.go
+++ b/internal/models/terminal.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"encoding/json"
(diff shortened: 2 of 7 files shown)


xarmian:
Thanks @asjdf for this contribution — and a well-built one: honoring per-collection terminal_options on both the CLI and server paths, keeping the KEEP-IN-SYNC contract between the two copies, and pinning each facet of the fix with its own test made this straightforward to review and merge. The review surfaced two narrow pre-existing edges (guest item-grant scoping in the grouped query, and the display layer hardcoding `status`) — both tracked internally as follow-ups, neither a fault of this PR. Hope to see you around the repo again.

PR #1182, opened 2026-08-23
fix(events): guard the generation counter the way the epoch key beside it is guarded (BUG-2740)

xarmian:
Closes BUG-2740.

## The unit

Every branch of `publishScript` that rotates the id space INCRs `pad:event_epoch_gen`, and none checked the key first — while the **epoch key beside it does**, with a comment saying exactly why.

An INCR against a corrupted key **aborts the script, after the sequence INCR has already landed**. Redis is atomic against interleaving but does not roll back a script's earlier writes, so the failure burns an id (a hole to every receiver), repeats on the next publish, and never self-heals: the branch that would rotate the generation is the branch that cannot run.

## Four abort modes, not the two the filing named

Measured against the pinned miniredis rather than assumed:

```
list                          -> WRONGTYPE
hash                          -> WRONGTYPE
string 'abc'                  -> ERR value is not an integer or out of range
string '9223372036854775807'  -> ERR increment or decrement would overflow
```

A TYPE check alone — the fix the filing describes — covers half. The value is validated on the same terms the epoch guard uses.

## Three things the review loop changed, each measured

**The ceiling is 17 digits, derived from the epoch guard's 18.** The value accepted here is about to be incremented, and the result becomes the *epoch*. Accept 18 and a counter at `999999999999999999` increments into a 19-digit epoch, that guard fires, and the script rotates a **second time inside one publish** — publishing a generation far below the one receivers hold.

**The generation is read back with GET, not stringified.** Redis hands an integer reply to Lua as a NUMBER, and Lua 5.1 doubles print with `%.14g`, so `tostring` diverges from the stored value well below the ceiling — measured at `99999999999999998`, which increments to `99999999999999999` in the key while `tostring` renders `100000000000000000`. Divergence begins at 2^53, not at the digit limit.

**A broken host clock is no longer fatal.** The seed is an input; an unset clock reports zero or a negative second, which the repair wrote to the key and returned as the epoch. The epoch guard rejects it, rotates, calls back into the repair, receives the *same* bad seed, and assigns it **without revalidating** — an unparseable epoch on the wire, rejected by every receiver. `clampGenerationSeed` forces a positive integer of at most 17 digits; the floor is 1 because a broken clock cannot deliver the ruling's property at all, and the real choice is between a value that is merely LOW (detected, one resync) and one that is FATAL.

## The ruling, and the residual it accepted

Restart at wall-clock seconds is **Dave's day-49 ruling**, read from the item's trail. Generations only need to be orderable among themselves, and the corrupted key cannot testify to the previous one.

It is not monotonicity. Two repairs inside one second seed the same value, so two id spaces can share an epoch — and an equal epoch means "same space" by design, so no epoch-based reason fires. **It is still not silent**, and the chain is now tested end to end rather than assumed:

> A merge requires ids REUSED at a receiver. Reuse requires the sequence to go BACKWARDS. A backwards counter is detected regardless of what the epoch says — `counter_backward`, which drops the affected buffers and refuses cursors below the discarded high-water mark.

The lead re-ruled on that probed fact: residual accepted **because it is detected**, with the attribution corrected from `epoch_regressed` to `counter_backward`.

Also recorded, after two reviews raised it: why a per-workspace backward check suffices for a global counter. The unit of coherence is the per-workspace buffer and cursor; a reissued id landing in a different workspace collides with nothing. Both shapes measured — a DELETED counter restarts at 1 and rotates the epoch unconditionally; a counter SET backwards above 1 skips that, and then only the workspace whose own ids are reissued needs a reset, and it gets one.

## Operator-facing

`docs/deployment.md` gains *A repaired generation counter*: a repair can surface as `epoch_change` or, if a collision pushed the counter higher, as `epoch_regressed` — which otherwise means a failover that lost writes. **There is no repair-specific counter or log**, because the repair happens inside Lua; the tell is the value, since a repaired generation looks like a unix timestamp. The exported Prometheus `HELP` and the metrics table say so too.

## Filed rather than fixed

**BUG-2744** — the sequence id is concatenated through the same Lua double. Reached by corruption rather than counting: a hand-edited or collided `event_seq` *arrives* at that magnitude exactly as the generation key does. Not fixed here because `assignScript`'s id IS consumed Go-side, so the remedy spans both paths.

## Gates

`make lint` 0 issues · `go test ./...` green · `-race` green on `internal/events` · `gofmt` clean. No store SQL and no web changes.

## Review

9 codex rounds, converged on the criterion: 6 and 9 CLEAN on broad angles, 8's finding refuted by probe. **17 mutations applied and grep-verified, 17 caught** — four only after a first version survived and exposed a test asserting an end state both the right and wrong code produce, or a table clearing state so uniformly it only ever drove one branch.

https://claude.ai/code/session_01JVDBKbgn3Xt7ndW1YoYd8X

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -412,13 +412,13 @@ Alert on these instead:
 | `pad_event_resume_gaps_total` | The ACTIVITY stream's (`/api/v1/events`) twin of the watch resume counter above. **Expect a step around a deploy, with the RATE settling back to baseline** (the counter itself only ever increases) — each instance starts with no replay coverage, so an early resume against a workspace it has not seen yet is a warranted resync. It counts RESUMES, not clients: a deploy with no reconnects does not move it at all, and a client that reconnects several times is counted several times. A rate that does not settle is the thing to alert on |
 | `pad_event_midstream_resyncs_total` | Activity-stream subscribers told MID-STREAM that they missed events, on a connection that stayed open. New in BUG-2730, and the counter to watch when judging whether that fix is costing more resyncs than it is worth. It counts ANNOUNCEMENTS, not causes and not distinct clients: a reset that drops buffers moves it once per live subscriber (and that ratio against `pad_event_sequence_resets_total` is the fan-out); a burst of drops on ONE connection moves it once, because signals coalesce and are rate-limited per connection; and a coverage loss on a workspace with no buffer yet moves it while every cause counter stays flat, because there was no coverage to end but the subscribers still have a hole |
 | `pad_watchevents_midstream_resyncs_total` (see also, listed above) | Same meaning for the watch stream. Its causes are a slow-subscriber drop and a received sequence gap or reset; a gap announces to EVERY subscriber on the instance, so it can exceed all of its cause counters |
-| `pad_event_sequence_resets_total` | Activity replay coverage dropped, by reason. `subscription_resumed` — a pub/sub connection dropped and resubscribed, dropping that workspace's buffer; expect it during a Redis failover and expect it to stop afterwards. `epoch_change` — the shared counter's ID space changed generation, dropping every buffer; expect a handful per cutover. `counter_backward` — an ID arrived at or below a buffer's high-water mark with no generation change; see *Event ID-space migration* for what to expect per phase. `epoch_regressed` — a LOWER generation was seen, so this instance stopped vouching for its buffers. One alongside an `epoch_change` is a message that was in flight when the generation rotated; a RUN of them means the counter itself went backwards, i.e. Redis lost writes. `undecodable_message` — a message on these channels could not be parsed, so that workspace's coverage ended; expect zero, and suspect a namespace collision |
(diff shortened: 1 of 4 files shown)


xarmian:
HELD behind BUG-2742, not abandoned — recording it here so the state is visible on the PR itself.

Three CI runs on this branch have gone red on `internal/events` timing tests. Each time a **different pre-existing test**, and on the third run a different **leg** (Go passed, Nix failed, having been the other way round the run before). None of the failures are in code this PR touches.

Established: unmodified `origin/main` fails the same way under local CPU contention. Package duration is identical between this branch and main (4.17s vs 4.13s), so it is not added runtime. The failing test's arrival count varies continuously — 64, 65, 150 of 320 — which is a throughput race, not a stuck consumer.

Per the lead's ruling: one CI cycle on a changed head is a test; a second on the same head would be roulette. So this holds until BUG-2742's family is fixed, then merges on a green run.

The change itself is complete: local gates green, 9 review rounds converged, 20 mutations applied and caught.

PR #1237, opened 2026-09-02
feat(cli): pad token create/list/revoke — CLI mint path for API tokens (#879 follow-up)

b4rk13:
## What does this PR do?

The follow-up invited in the #879 acceptance (grounding note 3): token minting was web-only, so the `PAD_TOKEN` override (#1160) wasn't usable end-to-end on a headless box — an agent's identity had to be minted in a browser. This adds a `pad token` command group over the **existing** user-scoped endpoints (`GET/POST /api/v1/auth/tokens`, `DELETE /api/v1/auth/tokens/{tokenID}`). No server changes.

Relates to #879 (deliberately doesn't close it — layer 2, named profiles, is still open).

- **`pad token create --name <name> [--expires-in <days>] [--scopes <s>]`** — mints and prints the secret exactly once, bolded on its own line, with a store-it-now notice; metadata (id, prefix, expiry) above it. `--name` is required locally (cobra `MarkFlagRequired`) so a nameless mint never reaches the server just to be refused. Reuses the TASK-788 structured plan-limit marker on `plan_limit_exceeded`, same as webhook create.
- **`pad token list`** — id, name, prefix, created, last-used, expires via tabwriter; metadata only, never secret material (the server doesn't return secrets on list, and the renderer can't invent one).
- **`pad token revoke <token-id>`** — exact id only, no prefix matching: a typo is a not-found error surfaced from the server's 404, never a wrong token revoked and never a claimed success.

`--format json` works on all three (create returns the full `APITokenWithSecret`).

Client methods live in a new `internal/cli/client_tokens.go` (the `client_sessions.go` pattern); commands in `cmd/pad/cmd_token.go`, wired as a top-level group since tokens are user-scoped, not workspace-scoped.

**Two flagged questions rather than assumptions:**

1. **MCP catalog: deliberately NOT exposed.** Minting/revoking credentials from an agent surface is what the catalog's exclusion rule ("destructive: mutates auth") exists for — an agent that can mint tokens can mint itself persistence. If you want `list` (read-only) on the catalog anyway, say so and I'll add just that.
2. **`rotate` exists server-side** (`POST /auth/tokens/{id}/rotate`) but isn't in this PR — kept to the minimal create/list/revoke scope. Happy to add `pad token rotate <id> [--expires-in]` here or as a follow-up, whichever you prefer.

## How to test

1. `PAD_TOKEN= pad auth login`, then `pad token create --name test-agent` — metadata plus the `pad_...` secret printed once, with the notice.
2. `pad token list` — the new token's row shows metadata and no secret.
3. `PAD_TOKEN=<the secret> pad auth whoami` — authenticates as the minting user (the #1160 override path, now end-to-end from the CLI).
4. `pad token revoke <id>` then repeat step 3 — the revoked token fails on its next call. `pad token revoke not-a-real-id` → not-found error, exit non-zero.

## Checklist

- [x] `go build ./...` passes (no web changes; `make build`'s web half untouched)
- [x] `go test ./cmd/pad/ ./internal/cli/` — zero new failures by name vs a clean-`main` baseline run on the same Windows box (the pre-existing HOME-assumption set, unchanged; new `TestToken*` all pass)
- [x] New features have tests — written first and watched fail (`undefined: tokenCreateCmd` at compile, then each assertion against the stub), driving the real commands with `HOME`+`USERPROFILE` both sandboxed per the #1160 convention
- [x] CLI help text updated — command/group help + README `PAD_TOKEN` section now documents the CLI mint path

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,13 +553,24 @@ Once a user exists, all API requests and web UI access require authentication. C
 
 #### Authenticating with an environment token
 
-Set `PAD_TOKEN` to a Pad API token (minted under **Settings → API tokens** in the web UI) to authenticate without `pad auth login`:
+Set `PAD_TOKEN` to a Pad API token (minted with `pad token create`, or under **Settings → API tokens** in the web UI) to authenticate without `pad auth login`:
 
 ```bash
 PAD_TOKEN=pad_xxxxxxxx pad item list
 ```
 
-`PAD_TOKEN` takes precedence over credentials saved by `pad auth login` — the same convention as `gh`'s `GH_TOKEN`. This is useful for CI, scripts, and machines where several AI agents share one CLI install but should act as different Pad users: give each agent its own token in its process environment, and the credential store is never touched. `pad auth whoami` reports the token's identity (with an `Auth: PAD_TOKEN environment override` line), and `pad auth login`/`logout` warn when the override is active — they manage the stored credentials, which the override bypasses. Deliberately, `pad auth logout` never invalidates the `PAD_TOKEN` session itself: it signs out the *stored* session only, and the env token's lifecycle belongs to wherever it was minted (revoke it under **Settings → API tokens**).
+`PAD_TOKEN` takes precedence over credentials saved by `pad auth login` — the same convention as `gh`'s `GH_TOKEN`. This is useful for CI, scripts, and machines where several AI agents share one CLI install but should act as different Pad users: give each agent its own token in its process environment, and the credential store is never touched. `pad auth whoami` reports the token's identity (with an `Auth: PAD_TOKEN environment override` line), and `pad auth login`/`logout` warn when the override is active — they manage the stored credentials, which the override bypasses. Deliberately, `pad auth logout` never invalidates the `PAD_TOKEN` session itself: it signs out the *stored* session only, and the env token's lifecycle belongs to wherever it was minted (revoke it with `pad token revoke` or under **Settings → API tokens**).
+
+#### Managing API tokens from the CLI
+
+```bash
+pad token create --name ci-agent              # Mint a token (secret shown once)
+pad token create --name cursor --expires-in 30
+pad token list                                # Metadata only — never secrets
+pad token revoke <token-id>                   # Immediate; the id must be exact
+```
+
+Tokens are user-scoped and act as the user who minted them. `create` prints the secret exactly once — the server stores only a hash and cannot show it again — so pair each mint with wherever the token will live (CI secret store, an agent's `PAD_TOKEN`). `revoke` takes the exact id from `pad token list`; revocation is immediate, and anything still authenticating with that token fails on its next call.
 
 ```bash
 pad workspace members               # List workspace members
--- a/cmd/pad/cmd_token.go
+++ b/cmd/pad/cmd_token.go
@@ -1,0 +1,187 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"text/tabwriter"
+	"time"
+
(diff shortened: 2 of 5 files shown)


xarmian (changes requested):
Thanks for this. It's a clean cut: it rides the existing `/auth/tokens` endpoints with no server change, `--name` is required client-side so a nameless mint never hits the server, revoke takes the exact id only (no prefix matching, so a typo is a 404 and never a wrong token gone), and the secret prints exactly once with the test asserting the count rather than presence. The MCP exclusion reasoning ("an agent that can mint tokens can mint itself persistence") is exactly right and I'm glad you did it unprompted.

One real problem, then answers to your two questions.

**`--scopes` mints dead tokens.** The flag passes its string straight through to `APITokenCreate.Scopes`. The store writes it verbatim (`internal/store/api_tokens.go:36-39` only substitutes `["*"]` for empty), and `tokenScopeAllows` (`internal/server/middleware_auth.go:1181`) requires a JSON array and denies on unparseable JSON. So `pad token create --name x --scopes read` stores the string `read`, and every request with that token is refused with a server-side slog warning the CLI user never sees. The web UI never sends scopes, so this PR is the first door that exposes it. Either fix is fine by me:

- (a) drop `--scopes` from this PR and let it ride with `rotate`, or
- (b) accept comma-separated values, validate against the known vocabulary (`*`, `read`, `write`, `pad:read`, `pad:write`, `pad:admin`), JSON-encode, and add one test that `--scopes read` posts `["read"]`.

Either way the help text "optional scopes string" should go.

Minor, non-blocking: `TestTokenList_RendersMetadataWithoutSecret` asserts `!strings.Contains(out, "secret")`, but the list stub never carries a secret, so that guard can't fail for the reason it names. Fold into whichever fix you pick if you're touching the test file anyway.

**Your questions.**

1. Read-only `token list` on the MCP catalog: no, not for now. Nothing asks for it yet, and it adds auth surface before we've settled a related question (below).
2. `rotate`: follow-up PR, please. It exists server-side, but its semantics (does the old secret die immediately, is there a grace window) deserve their own review, and this one is nicer single-purpose.

**Related, not on you:** the server currently lets a write-scoped PAT mint further PATs (`handleCreateUserToken` only checks for a user id, and PAT auth supplies one). Pre-existing, but this PR makes it a one-liner (`PAD_TOKEN=x pad token create`), which is the same persistence concern you named for MCP. I'm filing that as a separate issue to require session auth on token creation; it doesn't need to hold this PR.

Once `--scopes` is sorted I'll merge.

xarmian:
Merged as f262449b — thank you. Clean sit on the existing endpoints, and the exact-id-only revoke with its test is the detail that made this easy to take. One note for anyone reading later: since #1267, `pad token create` needs a login session (`pad auth login`); under `PAD_TOKEN` the server answers 403 session_required, while list/revoke keep working with a token. We'll add that line to the README.

b4rk13:
Thanks for taking it, and apologies for the quiet week — your review landed between working sessions and I only caught it post-merge. That leaves the `--scopes` finding live on main: as merged the flag still passes its string straight through, and nothing server-side validates it since. I'd like to make that right as your option (b) in a follow-up PR — comma-separated values validated against the known vocabulary, JSON-encoded, with the test that `--scopes read` posts `["read"]` — folding in the `TestTokenList_RendersMetadataWithoutSecret` guard you flagged. And I'll open `pad token rotate` separately as requested; my read of `RotateAPIToken` is the old secret dies immediately (hash replaced in place, no grace window), so the CLI wording would say so — flag me if you'd rather design a grace window first.

Issue #1221, opened 2026-08-30
CLI: atomic task claim/checkout to prevent concurrent double-work (execution lease)


b4rk13:
## Problem

Two processes that both read an item's status and conclude "no one else has started this" have no way to make that decision atomically. Whatever they do next — a status flip, a note, an external ledger entry — happens *after* the read, so a second process polling in the same window sees the same "unclaimed" state and proceeds too. Both make progress believing they hold exclusive execution rights; whichever writes back last wins, silently.

This isn't a hypothetical race. It shows up in exactly the setup CLI automation increasingly runs: a scheduled/cron-triggered process and a longer-running interactive process both driving the same board. Their timing is uncorrelated by construction — the cron process's poll interval and the interactive process's actual work duration have no relationship to each other — so a poll can land in precisely the window where an item has been claimed in some out-of-band way (a comment, a linked doc, an in-memory flag) but that claim isn't yet, and may never be, visible to anything reading the item itself. A restart or hand-off of the process holding the item widens the window further: if the "lock" lives only in that process's own state, it evaporates the moment the process does, and nothing tells a second reader the item was ever spoken for.

The only mitigation available today is convention — "only claim items in state X", "one runner per board", "treat backlog as an implicit lock" — enforced by whoever writes the polling logic, not by pad. Convention degrades exactly when it's needed most: under concurrency, restarts, and retries, which is when two readers are most likely to observe the same "unclaimed" snapshot.

pad already has a related but distinct mechanism: `update` accepts `expected_updated_at` for optimistic concurrency and returns a structured `update_conflict` 409 when the timestamp doesn't match (`internal/server/handlers_items.go`). That protects against *lost writes to a field* — I set a value expecting the item hasn't changed since I last read it. It doesn't protect against *concurrent claims for exclusive execution*, which is a different shape of problem: a claim is a right the caller wants to **acquire and later release**, not a value it wants to set once, and nothing today models "no one else is currently working on this" as a first-class, time-bounded state.

## Proposed fix

Add a claim/lease primitive independent of arbitrary field updates:

```
pad item claim <ref> --holder <id> [--ttl <duration>]
```

- Succeeds only if the item is currently unclaimed (or the caller already holds the live lease) and returns a lease: holder id, acquired-at timestamp, expiry. Default `--ttl` should be long enough to cover a typical automated run and short enough that a crashed holder doesn't block everyone else indefinitely (something in the 10-30 minute range as a starting default, tunable per call).
[...]

PR #1238, opened 2026-09-02
feat(items): execution lease — atomic claim/checkout with expiry (#1221)

b4rk13:
## What does this PR do?

Draft PR for #1221, built to the shape you approved and the two decisions pinned in [the thread](https://github.com/PerpetualSoftware/pad/issues/1221#issuecomment-5516497935) (default TTL 15 minutes; `Lease:` line on `show` / glyph marker on `list` / full object in JSON). Relates to #1221 — leaving close-on-merge to you.

**The primitive.** `POST .../items/{ref}/claim` and `.../release`, with `pad item claim <ref> [--holder <id>] [--ttl <dur>]` / `pad item release <ref>` in front. The claim is a conditional UPDATE whose predicate is the arbiter — deliberately the same protocol as the event-outbox claim (migration 083 / TASK-2714) and BUG-2415's orphan GC, dialect-uniform, no `FOR UPDATE SKIP LOCKED` special case. A store-level test races 8 goroutines and asserts exactly one winner.

- **Contention** answers `409 code=lease_held` naming the live holder, `acquired_at`, and `expires_at` — same envelope discipline as `update_conflict`, so a loser can log who won and skip without a second call.
- **Refresh:** a re-claim by the live holder moves the expiry forward and keeps `acquired_at` (heartbeat) — so long runs extend instead of everyone inflating the default TTL.
- **Expiry is the reaper:** an expired lease reads as absent on every path (claim predicate, item GET, list decoration) with no sweep job — test-pinned including the "crashed holder, no reaper ran" case.
- **Release** is idempotent (absent/expired → `released:false`, exit 0); releasing another holder's *live* lease is refused with the same 409.
- **Holder** defaults to the authenticated user's email (then id) — the #879 tie-in; freeform string until named profiles land, exactly as discussed.
- **TTL:** default 15m, per-call `ttl_seconds`, refused outside (0, 24h].

**Display**, per the thread: `Lease:` line under `Assigned:` in `PrintItemMeta` (absent when none); leased `list` rows get a cyan `»` on the ref/title cell via the existing pinned-`*` pattern — no new column, since `renderItemTable`'s width budget would tax every row's title for a usually-absent state; `"lease": {holder, acquired_at, expires_at}` in JSON on get and list, key omitted when absent (one point-read on GET, one workspace query on list — the shared item scan's column lists are untouched).

**Deliberate non-wirings, each open to your call:**

1. **No `updated_at` bump, no version entry** — a lease is coordination state, not content; a claim must never 409 a concurrent editor's `expected_updated_at` token.
2. **No SSE/activity event** — a heartbeat re-claim every few minutes would spam the feed; if you want live web-UI lease indicators, I'd emit claim/release (not refresh) events as a follow-up.
3. **No MCP catalog exposure** — that's a ToolSurfaceVersion decision and agents are the motivating users, so I'd rather add it to the catalog with your blessing than smuggle it in; say the word and it lands here or as a follow-up.
4. **No web UI rendering** — the TS `Item` type gained the optional `lease` field, but no component consumes it yet.

**Your codebase's own guards caught and shaped three things**, which was pleasing to watch: `TestNULColumnCensus` demanded a ruling on the three new columns (`lease_holder` is caller text → census entry + regenerated 084 triggers; the two timestamps are server-composed RFC3339 → recorded exclusions with reasons), `TestEveryRequestBodyReaderIsAccountedFor` demanded the new `decodeJSON` reader be ledgered (it is — same shape as `handlers_watches.go`), and `TestScanCostFiguresMatchTheList` kept `ScanNUL`'s cost figure honest (131→132).

**Postgres note:** `pgmigrations/062_item_lease.sql` mirrors the SQLite migration; I don't have a PG instance wired locally, so the PG path is written to the dialect-uniform pattern but only SQLite-tested here.

## How to test

1. Terminal A: `pad item claim TASK-5 --holder runner-a` → lease with holder + expiry. Terminal B: `pad item claim TASK-5 --holder runner-b` → exit non-zero, error names runner-a and the expiry.
2. `pad item show TASK-5` → `Lease: runner-a (expires in 14m)` under `Assigned:`; `pad item list` marks the row with `»`; `--format json` carries the lease object.
3. `pad item release TASK-5 --holder runner-a` → released; re-run → "No live lease" no-op, exit 0. B claims successfully now.
4. `pad item claim TASK-5 --ttl 5s`, wait 6s, claim as someone else → succeeds with no reaper.

## Checklist

- [x] `go build ./...` passes
- [x] `go test ./internal/store/ ./internal/server/ ./internal/cli/ ./cmd/pad/` — zero new failures by name vs a clean-`main` baseline run on the same Windows box (server and the new suites fully green; the pre-existing Windows HOME-assumption set unchanged)
- [x] New features have tests — store/server/CLI suites written first and watched fail (route 404s, `undefined:` compile errors); the display tests were written after their code, so I verified they bite by deliberately breaking the renderer and watching them fail before reverting
- [x] TypeScript types updated — `ItemLease` + optional `Item.lease`
- [x] CLI help text updated — command help + README (`pad item claim`/`release` rows)

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,6 +489,8 @@ pad item deps <ref>                   Show dependencies
 pad item unblock <src> <target>       Remove dependency
 pad item related <ref>                Show direct relationships for an item
 pad item implemented-by <ref>         Show incoming implementers for an item
+pad item claim <ref>                  Atomically claim an item for execution (--holder, --ttl; 409 names the live holder)
+pad item release <ref>                Release your execution lease (idempotent)
 pad item bulk-update --status X       Batch update multiple items
 
 pad collection list                   List collections with item counts
--- a/cmd/pad/cmd_item_lease.go
+++ b/cmd/pad/cmd_item_lease.go
@@ -1,0 +1,116 @@
+package main
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/fatih/color"
+	"github.com/spf13/cobra"
+
+	"github.com/PerpetualSoftware/pad/internal/cli"
+)
+
+// `pad item claim` / `pad item release` (#1221): the execution lease. A
+// claim answers "may I be the one executing this right now" atomically —
+// two pollers that both read "unclaimed" get one winner and one
+// structured refusal naming the live holder, instead of two winners.
+
+func itemClaimCmd() *cobra.Command {
+	var (
+		holderFlag string
+		ttlFlag    string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "claim <ref>",
+		Short: "Atomically claim an item for execution (lease with expiry)",
+		Long: `Acquire the execution lease on an item, or refresh it if you already
+hold it (a re-claim by the live holder extends the expiry — heartbeat).
+
+Fails with the live holder and expiry when someone else holds the item, so
+the loser can log who won and skip instead of double-working. The lease
+expires on its own; a crashed holder blocks nobody past the TTL.
+
+Examples:
+  pad item claim TASK-5
+  pad item claim TASK-5 --holder sweep-runner --ttl 30m`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ttlSeconds := 0
+			if ttlFlag != "" {
+				d, err := time.ParseDuration(ttlFlag)
+				if err != nil {
+					return fmt.Errorf("invalid --ttl %q: %w (use Go durations, e.g. 15m, 1h)", ttlFlag, err)
+				}
+				if d <= 0 {
+					return fmt.Errorf("--ttl must be positive, got %s", d)
+				}
+				ttlSeconds = int(d.Seconds())
+			}
+
+			client, _ := getClient()
+			ws := getWorkspace()
+
+			lease, err := client.ClaimItem(ws, args[0], holderFlag, ttlSeconds)
+			if err != nil {
+				return err
+			}
+
+			if formatFlag == "json" {
+				return cli.PrintJSON(lease)
+			}
+
+			green := color.New(color.FgGreen)
+			fmt.Printf("%s Lease acquired on %s\n", green.Sprint("✓"), args[0])
+			fmt.Printf("  Holder:  %s\n", lease.Holder)
+			fmt.Printf("  Expires: %s (%s)\n",
+				lease.ExpiresAt.Local().Format("2006-01-02 15:04:05"),
+				cli.LeaseCountdown(lease.ExpiresAt))
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&holderFlag, "holder", "", "lease holder identity (default: the authenticated user)")
+	cmd.Flags().StringVar(&ttlFlag, "ttl", "", "lease duration, e.g. 15m, 1h (default: 15m; max 24h)")
+
+	return cmd
+}
+
+func itemReleaseCmd() *cobra.Command {
+	var holderFlag string
+
+	cmd := &cobra.Command{
+		Use:   "release <ref>",
+		Short: "Release an item's execution lease (idempotent)",
+		Long: `Release the execution lease you hold on an item. Releasing an absent or
+already-expired lease is a no-op, not an error — cleanup code never needs
+to check whether it still holds the lease first. Releasing another
+holder's LIVE lease is refused.`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, _ := getClient()
+			ws := getWorkspace()
+
+			released, err := client.ReleaseItem(ws, args[0], holderFlag)
+			if err != nil {
+				return err
+			}
+
+			if formatFlag == "json" {
+				return cli.PrintJSON(map[string]any{"ref": args[0], "released": released})
+			}
+
+			green := color.New(color.FgGreen)
+			if released {
+				fmt.Printf("%s Lease released on %s\n", green.Sprint("✓"), args[0])
+			} else {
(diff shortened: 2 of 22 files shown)


b4rk13:
Gentle nudge on this one — it's sat as a draft since the 2nd while the token work went through, and main has moved ~900 commits and two releases underneath it.

The useful input from your side is the four calls flagged in the body (`updated_at`/version entry, SSE events, MCP catalog exposure, web-UI rendering) — or simply "shape's fine as drafted, finish it." Once I have that I'll rebase onto current main and mark it ready for review; the NUL-column census, the request-body-reader ledger, and the ScanNUL cost figures will all want re-satisfying after the drift, so I'd rather do that pass once, against your answers, than repeatedly.

And if it's parked on your side for capacity or sequencing reasons, that's a fine answer too — just say so and I'll leave it be until you call for it.

xarmian:
Very sorry for the silence. Shape's fine as drafted; here are the four calls:

1. updated_at / version entry: keep them out. A lease is coordination state, and a claim must never 409 someone's expected_updated_at. Reminders (#1244) took the same line.
2. SSE / activity events: keep them out of this PR. The structured 409 is the signal orchestrators need. If live indicators are wanted later, claim/release (not refresh) would ride the existing outbox rail like item.reminder_due.
3. MCP catalog: yes, please land it here. Add claim and release as pad_item actions in internal/mcp/catalog_item.go — additive, so no ToolSurfaceVersion bump (the file's DR-7 note covers it). Reminders added remind / ack-reminder the same way.
4. Web UI: not needed from you. The optional lease type on Item is enough; we'll add a badge on our side after merge.

Two mechanical notes for the rebase: main has migrations through 088 (SQLite) and 065 (Postgres), so yours become 089 and 066. Nothing touched nulcolumns.go, 084, or the ScanNUL figure since your branch, so the census / reader-ledger / cost re-satisfy should be quick.

Mark it ready when it's rebased and I'll take it to review. And thanks for extending the NUL census through the generator instead of by hand — that was the first thing checked and it was right.

b4rk13:
Rebased onto current main (`5557b306`, past v0.16.0) and marked ready. Your four calls, applied:

1. **`updated_at`/version: stays out** — as drafted; nothing changed.
2. **SSE/activity: stays out** — as drafted; the outbox-rail note stands if indicators are wanted later.
3. **MCP catalog: landed here.** `claim` + `release` as `pad_item` actions in `catalog_item.go`, additive per the reminders v0.28 disposition, no ToolSurfaceVersion bump (DR-7 grounds). stdio passes through to the CLI; HTTP gets a custom `mapItemClaim` — the catalog's `ttl` is a Go duration on both transports (the CLI flag's vocabulary), so the mapper owns the duration→`ttl_seconds` conversion and refuses a malformed or non-positive ttl rather than sending `ttl_seconds:0` for the server to default silently. `release` is a declarative routeSpec. Your guard suite named the rest of the duties and they're all satisfied: cmdhelp fixture, expected-cmdPath maps, `notFieldWriters` (`holder`/`ttl` are coordination state, not field writes), `instructions.md` bullet, README action table, and the parity test picks the routes up structurally. New `TestRoute_ItemClaim*` / `TestRoute_ItemRelease` cover mapper behavior.
4. **Web UI: nothing added** — the optional `lease` TS type only, as before.

Migrations renumbered `089`/`066` as you specified. **One consequence of the renumber you should rule on:** the draft had regenerated 084 with `items.lease_holder` triggers, and that's no longer tenable — the column now lands at 089, *after* the trigger migration, and a trigger referencing a not-yet-existent column breaks every later table-rebuild RENAME (086's `item_reminders` rename re-parses all triggers and fails with `no such column: NEW.lease_holder`; caught by your storetest fixtures the moment I rebased). So 084, `nulcolumns.go` and the ScanNUL figures are back to byte-identical with main, and the three lease columns are recorded in `nul_unprotected_baseline.txt` (via `GEN_NUL_BASELINE=1`) — the same ruling main applies to every post-084 table (`item_reminders`, `item_relation_links`). That does mean `lease_holder`, which is caller text, sits in the unprotected ledger rather than behind a trigger; if you'd rather it carried protection, say how you want post-084 columns to get triggers and I'll build to that — it looked like a repo-wide design question rather than something this PR should invent.

Verification: full `internal/store` / `internal/server` / `internal/cli` / `cmd/pad` / `internal/mcp` on this Windows box against a same-day clean-`main` worktree baseline — zero new failures by name (`TestDownload_BlobMissingOnDisk` flaked on *both* trees — a Windows file-lock race between the blob delete and the thumbnail decode, not this change; it passes in isolation on this branch 3/3 and failed on clean main 2/3). `go build ./...` and gofmt clean.

xarmian:
Merged. Thank you for this one — it's the cleanest outside contribution this repo has had.

Two things I want to name specifically. You read the NUL-invariant work that merged the same day you opened this and extended it correctly (the census entry, the regenerated 084 triggers, the ledgered reader), and then on the rebase you re-read the surface-versioning notes and asked instead of guessing. Both are things I'd have had to check for a contributor who didn't do them, and I didn't have to.

Three follow-ups, all ours, none of them yours to do:

1. The no-bump ruling I gave you on the catalog wiring was wrong. DR-7 is about a consumer tolerating an unknown key in the serialized surface, not about the action enum, and 0.28 bumped for exactly this shape. 0.38 lands in the next commit on main with a note saying so.
2. In the same commit: the mapper's `int(d.Seconds())` truncates, so a sub-second `ttl` was positive enough to clear the `<= 0` guard and still arrive as `ttl_seconds: 0`, which the server reads as absent and defaults to 15m. Now refused under 1s. One comparison; the guard's own comment describes the case.
3. The Layer B gap you raised in the commit message is real and repo-wide: every column added since 084 is outside it, because the generator can only rewrite 084. Tracked internally (BUG-3108); the fix is teaching the generator to emit incremental trigger migrations. Your read of why the alternative breaks the 086 rebuild was right.

#1221 closes with this.

PR #1271, opened 2026-09-07
fix(store): preserve valid item references across workspace import

### LINKED ISSUE #1270: Workspace import renumbers valid items and silently redirects existing references
## What happened?

Exporting and restoring a workspace can silently make an existing item reference identify a different task. `ImportWorkspace` assigns fresh consecutive `item_number` values even when the archive already contains valid workspace-unique numbers, but copies item content and comment bodies without rewriting numbered references.

A normal deletion is enough: deleting TASK-1 leaves TASK-2 and TASK-3. After export/import those rows become TASK-1 and TASK-2. Text that still says `[[TASK-2]]` now points at the former TASK-3. Export and import both succeed.

This affects workspace JSON and attachment-bundle imports and the SQLite→PostgreSQL migration command, which also calls `ExportWorkspace` followed by `ImportWorkspace`. It is reference corruption, not just cosmetic renumbering: instructions and commands using the original references address different items.

## Steps to reproduce

In a fresh workspace with an empty Tasks collection (prefix `TASK`):

1. Create TASK-1, titled `Obsolete task`.
2. Create TASK-2, titled `Deploy the release`.
3. Create TASK-3, titled `Release instructions`, with content `Follow [[TASK-2]] before continuing.`
4. Delete TASK-1, then export the workspace and import it under a new name.
5. In the imported workspace, resolve TASK-2 and inspect the instructions' content.

Expected: TASK-2 continues to identify `Deploy the release`.

Actual (repeated twice using the repository's SQLite store fixtures and real create/delete/export/import/resolve methods):

```text
before TASK-2="Deploy the release"
after TASK-2="Release instructions"
restored body="Follow [[TASK-2]] before continuing."
export/import succeeded
```

The same round trip without deletion passed twice. Fixture creation timestamps were fixed one second apart to remove same-second ordering as a confounding factor. Four existing numbering/export tests also passed.

## Component

API (shared store path; also affects CLI and web workspace imports).

## Environment

- Current upstream main: `e864a43644c8d694c45cd59a8ee5ccb5b1057908`
- The same unconditional renumbering is present in v0.15.0.
- macOS 26.6.2 arm64, Go 1.26.6, SQLite through the repository's normal store test fixture.
- No source modifications were needed for the reproduction; the test was injected through a Go overlay.

## Root cause and proposed scope

`internal/store/export.go:600–604,651,739` ignores the exported numbers and writes `nextItemNumber`. `internal/store/items.go:GetItemByRef` resolves by that number. `cmd/pad/cmd_db.go` uses the same importer for database migration.

#82 introduced renumbering to accept old per-collection exports that contain duplicate numbers. Keep that compatibility: preserve the exported numbers when the items that will actually be imported all have positive, workspace-unique numbers; retain the existing sequential fallback for legacy duplicate/missing/invalid numbers. No archive-format change or schema migration is needed.

Acceptance checks:

- A current archive with a deletion gap retains its reference targets and content/comment references.
- Valid numbers survive arbitrary archive order, including same-second exports.
- Creating an item after import allocates above the preserved maximum.
- Legacy duplicate or missing numbers still import with unique positive numbers.
- Skipped orphan items do not unnecessarily force renumbering of valid imported items.
- Run the regression against both SQLite and PostgreSQL.

## Related history

Repository-scoped searches covered all issue/PR states and result pages using `import`, `item_number`, `wiki export`, `renumber`, and `reference import`, followed by fresh `import number` and `renumber` searches. All three open PRs, including drafts, were inspected. No issue or active fixing PR for this defect was found. #82 explains the legacy fallback; #1255 and #1265 fix different export/import problems.

I will work on a focused fix and link the regression coverage in the PR.

mattfaltyn:
## What does this PR do?

Closes #1270.

After deleting TASK-1, importing a workspace containing TASK-2 and TASK-3 previously reassigned them to TASK-1 and TASK-2 while leaving `[[TASK-2]]` text unchanged. A successful restore could therefore redirect instructions to the wrong task; SQLite→PostgreSQL migration uses the same path.

Preserve exported item numbers when the items being imported have positive, workspace-unique numbers. Ignore skipped orphan rows when checking that condition. Retain the existing sequential fallback for legacy duplicate, missing, or invalid numbers, so old per-collection archives remain importable. No schema or archive-format changes, dependencies, or API-shape changes.

The regression covers the actual create/delete/export/import/resolve path, item and comment text, reordered same-timestamp archives, deletion gaps, allocation above the preserved maximum, legacy duplicate/missing/negative numbers, orphan rows, and empty archives. The backup guide documents preservation and the legacy fallback.

## How to test

```sh
go test ./internal/store -run '^TestImportWorkspace(PreservesReferenceTarget|ItemNumberCompatibility)$' -count=2 -v
PAD_TEST_POSTGRES_URL='postgres://.../pad?sslmode=disable' go test ./internal/store -run '^TestImportWorkspace(PreservesReferenceTarget|ItemNumberCompatibility)$' -count=2 -v
make build
make test
```

The deletion regression failed on upstream before the fix; the no-deletion control passed. Both pass after the fix. All validation below ran against commit `416dd404f43e81e4cd36c77ed928819ea08b5ad3`:

- Full Linux SQLite suite: 7,000 test/subtest passes, 63 repository-defined skips, 29 packages; repeated with `-race`, same result and no race reports.
- Full Linux PostgreSQL 17 suite: 7,008 test/subtest passes, 56 repository-defined skips, 29 packages; repeated with `-race`, same result and no race reports.
- Full Node 24 build, Tiptap pin check, svelte-check (0 errors; 6 existing warnings), and Vitest (2,189 tests across 134 files).
- Playwright: 204 passed, 192 repository-defined skips, no failures.
- `GOOS=linux make lint` (0 issues), `make vuln`, `npm run audit:ci`, and `git diff --check` passed.

The full backend commands were `go test -json -timeout=45m ./... -count=1` and `go test -race -json -timeout=45m ./... -count=1`, each with SQLite and `PAD_TEST_POSTGRES_URL` pointing at an isolated PostgreSQL 17 instance. These run the complete `make test` package set. Linux arm64 containers used Go 1.26.6 and RAM-backed test storage, matching the repository's test database setup. Full native SQLite and PostgreSQL store suites also passed.

Validation environment notes: the native macOS full suite encounters pre-existing session-owner/process-token failures, reproduced on unmodified upstream main, and a `/var` versus `/private/var` path assertion. The corresponding CLI packages pass on Linux. Playwright used an external launcher configuration to quote the checkout path's spaces and retain the original working directory; no test cases were changed or excluded.

Hosted workflows were approved while local validation was finishing. Web, Playwright, and macOS/Windows smoke checks have passed; Go, PostgreSQL, and Nix checks are still running as of the ready-for-review transition. The full local matrix above has completed successfully.

## Checklist

- [x] `make build` passes (Node 24, matching CI)
- [x] Complete `make test` suite passes on Linux, both databases, with and without `-race` (native macOS caveats above)
- [x] New features have tests (regression coverage for this fix)
- [x] TypeScript types updated (not applicable: no API-shape change)
- [x] CLI help text updated (not applicable: no new command)

### DIFF
--- a/docs/backup.md
+++ b/docs/backup.md
@@ -162,6 +162,13 @@ pad workspace import my-workspace.json
 pad workspace import --name "imported-workspace" my-workspace.json
 ```
 
+Item reference numbers are preserved when the imported items have positive,
+workspace-unique numbers, including gaps left by deleted items. This keeps
+references such as `[[TASK-2]]` pointing to the same task after restoration or
+SQLite→PostgreSQL migration. Legacy archives with duplicate, missing, or invalid
+numbers retain the sequential-renumbering fallback; references in those archives
+may need manual repair.
+
 ### One case where an export is not importable
 
 A workspace whose stored data contains a **NUL character** exports fine and is
--- a/internal/store/export.go
+++ b/internal/store/export.go
@@ -597,11 +597,25 @@ func (s *Store) ImportWorkspace(data *models.WorkspaceExport, newName string, ow
 		}
 	}
 
-	// Import items (first pass: create items, remap collection_id)
-	// Item numbers are assigned sequentially in created_at order to produce
-	// workspace-global numbering. Exported item_number values are ignored
-	// because old exports used per-collection numbering which can have
-	// duplicates within a workspace.
+	// Preserve modern archives' reference numbers, including deletion gaps and
+	// out-of-order items: content and comments still refer to those numbers.
+	// Legacy archives used per-collection numbering (or had no numbers), so
+	// retain sequential allocation when the imported set is not workspace-unique.
+	// Orphans are skipped below and must not force valid items to be renumbered.
+	preserveItemNumbers := true
+	seenItemNumbers := make(map[int]bool, len(data.Items))
+	for _, it := range data.Items {
+		if collMap[it.CollectionID] == "" {
+			continue
+		}
+		if it.ItemNumber <= 0 || seenItemNumbers[it.ItemNumber] {
+			preserveItemNumbers = false
+			break
+		}
+		seenItemNumbers[it.ItemNumber] = true
+	}
+
+	// Import items (first pass: create items, remap collection_id).
 	//
 	// IDEA-1486 + IDEA-1488: precompute each item's coerced fields/tags so
 	// the second-pass UPDATE (which re-applies fields after the ID remap)
@@ -649,6 +663,10 @@ func (s *Store) ImportWorkspace(data *models.WorkspaceExport, newName string, ow
 		parentID := resolveImportParent(it.ParentID, itemMap, insertedItems)
 
 		nextItemNumber++
+		itemNumber := nextItemNumber
+		if preserveItemNumbers {
+			itemNumber = it.ItemNumber
+		}
 		// IDEA-1486 + IDEA-1488: coerce empty-string / malformed
 		// fields/tags at the import boundary. After migration 056 /
 		// pgmigrations 035 hardened items.fields and items.tags to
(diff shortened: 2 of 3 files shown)


xarmian:
Thank you — this is a model contribution. You found the defect with a reproduction, root-caused it to the exact lines (`ImportWorkspace` reallocating `item_number` while content and comments keep the old references), checked #82's compatibility constraint before touching the fallback, and wrote the acceptance criteria including both drivers. I verified the mechanism on `main` independently, confirmed the pre-scan's skip predicate matches the import loop's orphan gate so a skipped item cannot force valid ones back onto sequential numbering, and confirmed later creates allocate above the preserved maximum under the workspace lock. Both the SQLite and PostgreSQL jobs ran green on your head.

One follow-up we are filing on our side, not a blocker: when an archive falls back to sequential renumbering (duplicate, missing or invalid numbers), it does so silently, and the other coercions in that function log when they rewrite something. A `slog.Warn` there would tell an operator why their `[[TASK-2]]` references moved. Happy to take that from you too if you want it.

Merging now; this will ship in the next release. Tracked internally as BUG-2918.

PR #1289, opened 2026-09-08
fix(cli): use PostgreSQL connection strings for backup and restore

### LINKED ISSUE #1288: PostgreSQL backup and restore treat PAD_DATABASE_URL as a literal database name
## What happened?

`pad db backup` and `pad db restore` cannot use a valid PostgreSQL connection URI supplied through the documented `PAD_DATABASE_URL` setting. Both wrappers set the child process's `PGDATABASE` to the entire URI without passing an explicit database argument. libpq treats that environment value as a literal database name, rather than expanding the URI into connection parameters.

This blocks both creating PostgreSQL backups and restoring existing archives through Pad. The commands fail explicitly; this report does not claim silent data loss.

## Steps to reproduce

Against an isolated PostgreSQL database with a valid connection URI:

1. Create a table containing one identifiable row.
2. Set `PAD_DATABASE_URL` to the URI. Run `pad db backup --output backup.sql`.
3. Produce a valid archive with `pg_dump --dbname "$PAD_DATABASE_URL" --format plain --clean --if-exists --file backup.sql`.
4. Set `PAD_DATABASE_URL` to a second disposable database's URI and run `pad db restore --force backup.sql`.
5. Compare with `psql --dbname "$PAD_DATABASE_URL" --file backup.sql --single-transaction --set ON_ERROR_STOP=on`.

Expected: Pad connects to the database named by the URI, creates a usable archive, and restores its rows.

Actual, reproduced twice per command using the real Pad binary and PostgreSQL clients:

```text
PAD BACKUP: exit=1
pg_dump: ... FATAL: database "postgresql:///source?host=..." does not exist
Error: pg_dump failed: exit status 1

PAD RESTORE: exit=1
psql: ... FATAL: database "postgresql:///target?host=..." does not exist
Error: psql restore failed: exit status 2
```

Native `pg_dump --dbname` succeeds with the identical source URI. Native `psql --dbname` succeeds twice with the identical destination URI and restores the expected row. Pad backup succeeds when given a bare database name with the same PG environment, isolating the URI interpretation mismatch. The reproduction used private Unix sockets and synthetic data; the ellipses above abbreviate temporary socket paths.

## Component

CLI: `cmd/pad/cmd_db.go`, `dbBackupCmd` (lines 124–132) and `dbRestoreCmd` (248–254). `docs/backup.md` documents the complete URI configuration, including the cron example.

## Environment

- Canonical main: `3d78e1d03e1f8226a6dc95aa761a93a23db31460`; both assignments also exist in v0.15.0.
- macOS 26.6.2 arm64; Go 1.26.6; PostgreSQL server/pg_dump/psql 18.4.
- Clean upstream code, built in an isolated checkout; no production database or external service involved.

## Proposed fix and acceptance criteria

Use a shared native-client connection handoff that lets libpq interpret the connection string explicitly. Preserve the existing intention to keep passwords out of process arguments, rather than moving a password-bearing URL directly to argv.

- Both backup and restore work with `postgres://` and `postgresql://` URIs and keyword/value connection strings.
- Preserve endpoint, database, user, escaped values, TLS/query options, and native connection semantics.
- Connection passwords do not appear in the subprocess arguments.
- Verify an actual backup→restore round trip and its row contents against PostgreSQL; include meaningful connection failure coverage.
- Preserve SQLite behavior; no schema migration or new dependency is needed.

## Related history

Rechecked all-state repository issue/PR searches, including comments and all pages, for `restore`, `psql`, `pg_dump`, `PGDATABASE`, and related symptom terms. No matching issue or active fixing PR found. #69 introduced the wrappers; #98 and #837 fix SQLite behavior and leave these PostgreSQL assignments unchanged. #843 concerns migration guards and SQLite snapshots.

I will submit a focused fix with regression coverage linked to this issue.

mattfaltyn:
## What does this PR do?

Closes #1288.

`pad db backup` and `pad db restore` passed the complete PostgreSQL URI through `PGDATABASE`, which native clients treat as a literal database name. Both commands now use one shared handoff with an explicit `--dbname` connection string and a separate password environment value, so documented connections work without placing database passwords in argv.

The helper preserves native connection options and quoted keyword values. Service credentials remain in the service configuration: inline password/service combinations and inline SSL key passphrases are refused with a specific remedy rather than changing password precedence or exposing them in argv. SQLite behavior is unchanged; no dependencies or schema changes.

## How to test

1. `go test ./cmd/pad -run '^TestPostgresClient' -count=1` covers URI/query and keyword escaping, password handoff, defaults, and invalid input.
2. With matching `pg_dump` and `psql` clients, set `PAD_TEST_POSTGRES_TOOLS_URL` to an isolated PostgreSQL server whose account can create databases, then run `go test ./cmd/pad -run '^TestPostgresBackupRestoreConnection$' -count=2`. The test creates and removes its own source/destination databases; it verifies row contents through real backup→restore round trips for both URI schemes and keyword strings, plus a missing-database failure.
3. Full local Linux SQLite and PostgreSQL suites passed, both normally and with the race detector: `go test -timeout=45m ./... -count=1` and `go test -race -timeout=45m ./... -count=1`, with `PAD_TEST_POSTGRES_URL` set for each PostgreSQL run. All four runs exited 0.

Additional validation: full build; native PostgreSQL 18.4 with SCRAM authentication and an escaped password (six successful round trips); frontend type checks (0 errors, 6 existing warnings); all 2,281 frontend unit tests; browser E2E (204 passed, 192 intentional skips); Linux golangci-lint v2.11.4 (0 issues). The live regression fails with the original command implementation and passes with this patch.

GitHub CI and Nix workflows currently require maintainer approval (`action_required`). Local full Go validation uses Linux/arm64 with Go 1.26.6; the native macOS lint run encountered the existing Linux-only `errProcStatMalformed` unused-symbol issue. No unrelated platform changes are included.

## Checklist

- [x] `make build` passes
- [x] `make test` equivalent passes (full Linux SQLite and PostgreSQL suites)
- [x] Regression tests added
- [x] TypeScript types updated if API changed — not applicable
- [x] CLI help text updated if new command — no new command; backup documentation updated

Validation environment: Go 1.26.6 on Linux/arm64 with PostgreSQL 18.6 for the full matrix; native macOS/arm64 with PostgreSQL 18.4 for the real backup/restore regression. Initial Linux runs were discarded after the VM ran out of disk space; all four complete suites were rerun successfully after expanding its disk. Browser E2E used the standard repository configuration with an isolated server port and a quoted absolute launcher path for the workspace path containing spaces. Native `make install` was replaced by the documented sibling-safe approach: build and exercise this worktree’s binary without stopping other Pad sessions.

### DIFF
--- a/cmd/pad/cmd_db.go
+++ b/cmd/pad/cmd_db.go
@@ -8,7 +8,6 @@ import (
 	"log/slog"
 	"net/url"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -128,8 +127,10 @@ the server resolves it (PAD_DB_PATH > PAD_DATA_DIR/pad.db > ~/.pad/pad.db).`,
 					"--file", output,
 				}
 
-				pgCmd := exec.Command("pg_dump", pgArgs...)
-				pgCmd.Env = append(os.Environ(), "PGDATABASE="+dbURL)
+				pgCmd, err := postgresClient("pg_dump", dbURL, pgArgs...)
+				if err != nil {
+					return err
+				}
 				pgCmd.Stdout = os.Stdout
 				pgCmd.Stderr = os.Stderr
 
@@ -250,8 +251,10 @@ WARNING: This will overwrite the current database contents.`,
 					"--single-transaction",
 				}
 
-				psqlCmd := exec.Command("psql", psqlArgs...)
-				psqlCmd.Env = append(os.Environ(), "PGDATABASE="+dbURL)
+				psqlCmd, err := postgresClient("psql", dbURL, psqlArgs...)
+				if err != nil {
+					return err
+				}
 				psqlCmd.Stdout = os.Stdout
 				psqlCmd.Stderr = os.Stderr
 
--- a/cmd/pad/postgres_client.go
+++ b/cmd/pad/postgres_client.go
@@ -1,0 +1,138 @@
+package main
+
+import (
+	"errors"
+	"net/url"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// postgresClient passes an explicit conninfo to libpq: PGDATABASE alone does
+// not expand a URI. Extract the database password so it stays off argv.
+func postgresClient(name, conninfo string, args ...string) (*exec.Cmd, error) {
+	dsn, password, err := postgresClientConninfo(conninfo)
+	if err != nil {
+		return nil, err
+	}
+	if password != nil && os.Getenv("PGSERVICE") != "" {
+		return nil, errors.New("configure the password in the PostgreSQL service instead of inline with PGSERVICE")
+	}
+	cmd := exec.Command(name, append(args, "--dbname", dsn)...)
+	cmd.Env = os.Environ()
+	if password != nil {
+		cmd.Env = append(cmd.Env, "PGPASSWORD="+*password)
+	}
+	return cmd, nil
+}
+
+func postgresClientConninfo(raw string) (string, *string, error) {
+	invalid := errors.New("invalid PostgreSQL connection string")
+	if strings.HasPrefix(raw, "postgres://") || strings.HasPrefix(raw, "postgresql://") {
+		u, err := url.Parse(raw)
+		if err != nil || u.Fragment != "" {
+			return "", nil, invalid // url.Error includes the supplied password.
+		}
+		var password *string
+		if u.User != nil {
+			if value, ok := u.User.Password(); ok && value != "" {
+				password = &value
+				u.User = url.User(u.User.Username())
+			}
+		}
+		// libpq uses percent decoding, NOT form decoding: '+' is literal.
+		var query []string
+		service := false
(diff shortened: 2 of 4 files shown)


xarmian:
Merged, thank you. Two in two days, and both were the kind of fix that reads as if the author has been bitten by the thing before.

The parts I want to call out, because they are exactly where this class of patch usually goes wrong: `url.PathUnescape` on the query tokens rather than `ParseQuery`, so `+` stays literal the way libpq actually reads it; non-password tokens re-joined from the original bytes, so multi-host URIs and quoted keyword values survive untouched; and refusing an inline password alongside a service file with a remedy, instead of silently demoting it under `PGPASSWORD`. That last one is a judgement call and I think you made the right one.

One thing on our side, not yours: the round-trip test can't run in CI yet. The service container is Postgres 17 and the runner ships client 16, so setting `PAD_TEST_POSTGRES_TOOLS_URL` there today would turn a skip into a red suite until we pin a matching client install. Filed internally as a follow-up we own.

Tracked as BUG-2952 on our side; #1288 closes with this.

PR #1303, opened 2026-09-09
fix(nix): gate the vendorHash heal on state, not on the push range (BUG-2974)

xarmian:
Closes the durable half of BUG-2974. The immediate half (main's stale hash) was fixed by the lead's direct commit `5ce11039`; this is the mechanism so it cannot recur.

## The defect

`push-vendor-hash` gated on the push RANGE — did *this* push touch `go.mod`/`go.sum`. Right question for loop prevention, wrong one for recovery: after a lost push race the tree that needs healing was written by an **earlier** push, so every later merge that did not itself move the module set refused to fix it and **exited 0**. Observed on `b7235a1a` → `bc543a1d`: the first run committed the heal and lost the push race, the second run recomputed the identical mismatch and skipped with `go.mod/go.sum unchanged across b7235a1a..bc543a1d`.

CI could not report it, and that is by design working correctly: the recompute repairs the hash in the **working tree** before the build steps judge it (TASK-2954's red-baseline fix), so `nix flake check`, `nix build`, the smoke test and the vulnerability scan all pass against a corrected tree while the committed file stays wrong.

## The fix

**Gate on STATE.** Does what main COMMITTED differ from what the build job RECOMPUTED? That question is equally true for a hash this push broke and one an earlier push broke and could not land, so the run after a lost race heals it.

**The loop guard never depended on the range.** The heal commit corrects `package.nix`, so the run it triggers finds the build green, never sets `bumped`, and the job does not start at all. The old comment named the range gate as the loop guard; it was not, and that mis-description is what made the gate look safe to rely on.

**An unhealed hash is LOUD.** A rejected push, and a recomputed artifact byte-identical to the committed file, both end the run RED with an `::error::` naming the state — instead of the silent `exit 0` that hid this. Property, as ruled on the bug's trail: *within one main run of any merge that moves the hash, either `nix/package.nix` on main carries the new hash or the run is red saying so.*

**The step is a script, not a `run:` block.** `nix/heal-vendor-hash.sh`, same shape as `nix/bump-vendor-hash.sh`. That is what lets it be tested — the defect shipped green for a day precisely because nothing outside a real merge race could exercise it.

## Verification

`nix/heal-vendor-hash_test.sh` — 18 assertions, driven against a real git remote, `git` the only dependency (~2s). Wired into the Go job beside the parser suite and into `make test-nix-hash`.

The load-bearing case is the one that shipped, and it carries a **frozen reproduction of the pre-fix gate as a negative control**: that leg exits 0 and leaves main stale, which is the bug, and it will keep doing so.

| case | asserts |
|---|---|
| 1 | heal owed by an EARLIER push, this push touched no `go.mod` — pre-fix form exits 0 and leaves main STALE; new form heals |
| 2 | the ordinary same-push heal still works (no regression) |
| 3 | artifact identical to the committed file → exit 1 + `::error::`, not a silent no-op |
| 4 | lost push race → exit 1, says the heal did not land, main untouched |
| 5 | the NEXT push to main then heals it with nobody intervening |
| 6 | the script refuses to run outside CI (it commits and pushes) |
| 7 | missing `SHA` fails before staging anything, naming the input |

**Mutation matrix — 6 mutants, all killed.** Contradiction arm exits 0 → case 3. Bare `git push` → case 4. Push-range gate re-introduced → 6 assertions. State comparison inverted → 8. Local-run guard removed → case 6. `SHA` requirement removed → survived at first (near-equivalent under `set -u`), so case 7 was strengthened to assert *where* it stops, not just the exit code; then killed.

Gates: `make test-nix-hash` green unpiped (40 parser + 18 heal). Both workflow files parse. No Go or web source changed.

## Also folded in

The two prose warts parked on TASK-2954's trail, per that trail's instruction to fold them into the next `nix.yml` touch:

- the doubled clause at `nix.yml:144` (*"self-heals one commit after the merge — usually one commit after it"*), and its over-long line;
- the comment naming the loop guard as the commit's own contents via the range gate — corrected to name the real terminator.

The second one turned out to be this bug rather than a wording problem, which is the reason it is worth saying explicitly in the file.

## Deliberately NOT here

**A rebase-and-retry on the rejected push.** It was listed on the bug as a shape worth weighing and it would make the losing run heal instead of deferring, but it needs an arm deciding whether the new tip moved the module set (our recomputed hash is only right if it did not), and neither arm can be exercised on this box. State-gating already makes the next push recover it, which is the property that was asked for. Worth a separate item if the deferral is ever felt.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,16 @@ jobs:
       - name: Test the Nix vendorHash bump parser
         run: nix/bump-vendor-hash_test.sh
 
+      # The heal job that pushes the parser's output to main (BUG-2974). Gated
+      # here for the same reason and one more: in nix.yml that job runs only
+      # after a real merge has moved the module set, so its defect shipped green
+      # for a day and was found by reading run logs. The suite drives the script
+      # against a real git remote, and its load-bearing case is the one that
+      # shipped — a heal owed by an EARLIER push, on a push that touched no
+      # go.mod/go.sum.
+      - name: Test the Nix vendorHash heal job
+        run: nix/heal-vendor-hash_test.sh
+
       - name: Run golangci-lint
         # only-new-issues: false means CI fails on ANY linter finding,
         # not just findings on PR-changed lines. The IDEA-732 cleanup
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -11,17 +11,21 @@ concurrency:
   # same PR makes the older run's answer worthless.
   #
   # PUSH runs get a group of their OWN, one per commit, and this is deliberate
-  # (TASK-2954). A run on `main` may owe main a vendorHash heal, and losing it
-  # loses the fix SILENTLY: the next push's run recomputes correctly, but its
-  # go.mod/go.sum gate sees only its own commits, so nothing ever retries the
-  # skipped heal. `cancel-in-progress: false` is not enough to prevent that —
+  # (TASK-2954). A run on `main` may owe main a vendorHash heal, and an evicted
+  # run never performs it. `cancel-in-progress: false` is not enough on its own —
   # GitHub holds only ONE pending run per group, so a third push evicts the
   # queued second, and the eviction looks exactly like a run that decided there
   # was nothing to do. A per-commit group cannot be evicted by anything.
   #
+  # What an evicted heal costs is now DELAY, not the fix. Since BUG-2974 the
+  # heal is gated on the committed hash rather than on what the push touched, so
+  # the next push to main meets the same stale hash and heals it from its own
+  # run. The per-commit group is what makes the heal land on the merge that
+  # caused it instead of waiting for whatever lands next.
+  #
   # What that trades for: two main runs can reach the push step at once, and the
   # loser gets a non-fast-forward rejection and goes red. That is the right
-  # direction — a red job is read, and the next merge's run heals anyway.
+  # direction — a red job is read, and the next push to main heals anyway.
   group: nix-${{ github.event_name == 'push' && github.sha || github.ref }}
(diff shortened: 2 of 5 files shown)


xarmian:
**Codex round 1 found a P1 in my own test, round 2 CLEAN. Fixed at `57128e77`.**

The finding: GitHub Actions always sets `GITHUB_ACTIONS=true`, and the test process inherits it — so case 6, written to prove the script refuses to run outside CI, satisfied the very guard it was tripping. It passed locally and would have **failed the new Go CI step on every run**. The suite's first act in CI would have been to report a defect it had introduced itself.

Reproduced before fixing (`GITHUB_ACTIONS=true nix/heal-vendor-hash_test.sh` → 16 passed, 2 failed), which is the check I owed and had not run: I verified the suite in the environment I was standing in, not the one it ships to.

**Fixed at the class, not the instance.** The suite now unsets `GITHUB_ACTIONS` and `HEAL_ALLOW_LOCAL` up front, so no case can be decided by the caller's environment, and each opts in explicitly. That also exposed an arm nothing asserted — the guard's POSITIVE side, where the real job runs in Actions with no override — now case 6b.

Green in four environments: bare, `GITHUB_ACTIONS=true`, a stray `HEAL_ALLOW_LOCAL=1`, and `init.defaultBranch=master`. Mutation matrix re-run and extended to seven (added: guard condition inverted); all seven killed.

**What I take from it, beyond the fix.** The whole argument for turning the `run:` block into a script was *testability* — and the test I wrote to justify that was itself untested against the environment it runs in. Same shape as the mutant that survived an hour earlier: I asserted the thing I was thinking about and not the condition around it. The instrument needs its own negative control, and for anything that ships to CI, "does it still hold with CI's environment set" is that control.

PR #1329, opened 2026-09-11
fix(cli): `pad token create --scopes` validates and JSON-encodes (#1237 review option b)

b4rk13:
## What does this PR do?

Closes out the one real finding from [the #1237 review](https://github.com/PerpetualSoftware/pad/pull/1237#pullrequestreview-5118132360), which merged before I got back to it: `--scopes` passed its raw string into `APITokenCreate.Scopes`, the store writes it verbatim, and `tokenScopeAllows` requires a JSON array — so `pad token create --name x --scopes read` minted a token every request refused, with only a server-side slog warning the CLI user never sees. This is your option (b):

- **Comma-separated values**, trimmed, **validated against the server's recognized vocabulary** (`*`, `read`, `write`, `pad:read`, `pad:write`, `pad:admin` — read from `tokenScopeAllows`, cited in a comment next to the CLI's copy of the list). An unknown or empty entry is refused locally with the allowed list named, before any request — same disposition as `--name`'s local requirement.
- **Valid values are JSON-encoded**: `--scopes read` posts `"scopes": "[\"read\"]"`, test-pinned against the stub server's recorded body (four cases including whitespace-padded multi-value and `*`).
- Help text no longer says "optional scopes string".

**Folded in, per the review's minor note:** `TestTokenList_RendersMetadataWithoutSecret`'s no-secret guard could not fail for the reason it named, because the list stub never carried a secret. The stub's first row now carries a `token` field the real server never sends on list, so the assertion is falsifiable: the renderer must keep printing known metadata columns only, even from a buggy or hostile server. Verified it bites by reintroducing exactly that leak (a `Token` field on `models.APIToken` plus a renderer column) and watching the test fail on the leaked value before reverting.

No server changes. The README's session-required section (#1267) already covers the other loose end from the merge comment, so nothing to add there.

## How to test

1. `pad token create --name scoped --scopes read` → token minted; a request with it succeeds on GET and is refused on POST (the scope finally does what it says).
2. `pad token create --name scoped --scopes read-only` → refused locally: `unknown scope "read-only"; allowed values: *, read, write, pad:read, pad:write, pad:admin`; no request sent.
3. `pad token create --name scoped` (no flag) → unchanged: no `scopes` key posted, server records full access.

## Checklist

- [x] `go build ./...` passes
- [x] `go test ./cmd/pad/` — zero new failures by name vs the same-day clean-`main` run on this Windows box (the pre-existing HOME-assumption set, unchanged; all `TestToken*` pass)
- [x] New features have tests — written first and watched fail against the pass-through (`posted scopes "read", want "[\"read\"]"`)
- [x] CLI help text updated

### DIFF
--- a/cmd/pad/cmd_token.go
+++ b/cmd/pad/cmd_token.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
 	"text/tabwriter"
 	"time"
 
@@ -54,11 +57,16 @@ func tokenCreateCmd() *cobra.Command {
 		Short: "Mint a new API token (secret shown once)",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			scopes, err := encodeTokenScopes(scopesFlag)
+			if err != nil {
+				return err
+			}
+
 			client, _ := getClient()
 
 			input := models.APITokenCreate{
 				Name:      nameFlag,
-				Scopes:    scopesFlag,
+				Scopes:    scopes,
 				ExpiresIn: expiresInFlag,
 			}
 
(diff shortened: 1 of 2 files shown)


xarmian:
Merged, thanks. This one was doing real damage quietly: a token created with `--scopes read,write` looked fine and then failed every request, and nothing in the flow said why. Validating against the server's own scope set and refusing locally with the allowed list is the right shape. Checked that the vocabulary matches tokenScopeAllows exactly before merging.

PR #1330, opened 2026-09-11
feat(cli): `pad token rotate` — new secret in place, old one dies with the write (#1237 follow-up)

b4rk13:
## What does this PR do?

The `rotate` follow-up you asked for [in the #1237 review](https://github.com/PerpetualSoftware/pad/pull/1237#pullrequestreview-5118132360) — `POST /auth/tokens/{id}/rotate` existed server-side with no CLI in front of it. Single-purpose, independent of the `--scopes` fix in #1329 (both branch off current main; whichever merges second has no conflict — they touch different regions of `cmd_token.go`, and I'll rebase if GitHub disagrees).

**On the semantics question your review raised** (does the old secret die immediately, is there a grace window): reading `RotateAPIToken`, the old secret dies with the same UPDATE — `token_hash`/`prefix` are replaced in place on the row, so there is no grace window and no window to design around in the CLI. This PR takes that as the contract and says it out loud rather than hedging:

- `pad token rotate <token-id> [--expires-in <days>]` — exact id only, same discipline as revoke: a typo is the server's 404 surfaced, never a wrong token rotated (which kills a live secret somewhere).
- The new secret prints exactly once (count-pinned test, same as create), with the store-it-now notice **plus "The old secret stopped working immediately."** — the wording is test-pinned so it can't drift into implying a grace window.
- `--expires-in` posts `{"expires_in": N}` (server-capped by max lifetime); without the flag no `expires_in` key is posted and the original expiry is preserved — test-pinned both ways.
- README documents `rotate` beside create/list/revoke, folded into the session-required section: rotate shares create's #1267 gate (`403 session_required` under a `pad_` PAT), which the README previously only quoted in the error text.

If you'd rather the server grow a grace window before this ships, happy to park it — the CLI wording is the only thing that would change.

## How to test

1. `pad token create --name ci` → use the secret in `PAD_TOKEN`, confirm a call works.
2. `pad token rotate <id>` (from a logged-in session) → new secret printed once; the old `PAD_TOKEN` value fails on its next call; the new one works. `pad token list` shows the same id/name with a new prefix and `last used: never`.
3. `pad token rotate <id> --expires-in 7` → expiry moves; without the flag it stays.
4. `pad token rotate not-a-real-id` → not-found error, exit non-zero.
5. `PAD_TOKEN=<pad_ token> pad token rotate <id>` → `403 session_required`, same as create.

## Checklist

- [x] `go build ./...` passes
- [x] `go test ./cmd/pad/ ./internal/cli/` — zero new failures by name vs the same-day clean-`main` run on this Windows box (the pre-existing HOME-assumption set, unchanged; all `TestToken*` pass)
- [x] New features have tests — written first and watched fail (`undefined: tokenRotateCmd`), driving the real command against the stub's recorded paths/bodies
- [ ] TypeScript types updated — n/a, no API change (the web settings page already rotates via its own client)
- [x] CLI help text updated — group example + command help + README


xarmian:
Merged, thanks. The rotate endpoint has been sitting on the server without a CLI door for a while, so this is a genuine gap-fill. Appreciated that the doc comments match what the handler does, including the 403 on non-session auth, which is already covered by a server test.

PR #1332, opened 2026-09-11
fix(cli): refuse SQLite self-restore

### LINKED ISSUE #1331: SQLite db restore truncates the database when source and destination are the same file
## What happened?

`pad db restore` accepts the resolved live SQLite database itself as the backup file. The command opens that file for reading, reopens the same inode with `os.Create`, truncates it to zero bytes, copies EOF, exits 0, and prints `Restore complete`.

This also affects aliases of the same file, including symlinks and hard links. The restore should reject any source that identifies the destination inode before opening the destination for writing.

## Steps to reproduce

Use a disposable temporary directory; this intentionally exercises a destructive bug.

```bash
repro_dir="$(mktemp -d)"
sqlite3 "$repro_dir/pad.db" \
  'PRAGMA user_version=1; CREATE TABLE proof(value TEXT); INSERT INTO proof VALUES("kept");'

shasum -a 256 "$repro_dir/pad.db"

env -u PAD_DB_DRIVER -u PAD_DATABASE_URL -u PAD_DB_PATH \
  PAD_DATA_DIR="$repro_dir" PAD_PORT=65434 \
  pad db restore --force "$repro_dir/pad.db"

echo "exit=$?"
stat -f '%z bytes' "$repro_dir/pad.db"
shasum -a 256 "$repro_dir/pad.db"
```

I repeated this against three independently created valid SQLite databases on current `main`: two 4,096-byte fixtures and one 8,192-byte fixture. Every run exited 0, printed a successful restore message, and reduced the database to zero bytes (SHA-256 `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`).

A control restore from a distinct `backup.db` to the resolved `pad.db` succeeded and produced a byte-identical destination.

## Expected behavior

If the backup and destination identify the same underlying file, restore should return a clear error before changing either file. Direct paths, symlink aliases, and hard-link aliases should all be rejected.

## Actual behavior

The command exits 0 and prints:

```text
Restoring SQLite database .../pad.db from .../pad.db...
Restore complete. Restart the Pad server to pick up the restored database.
```

The live database is now zero bytes.

## Impact

This can erase the complete live Pad database through a documented administrative command while falsely reporting success.

## Environment

- Component: CLI
- Pad: v0.15.1 and current `main` at `63adbebb76e2ef62f46e3c9d45a7f0525d514611`
- OS: macOS 26.6.2, Darwin arm64
- Go: 1.26.6
- SQLite: 3.51.0

## Root cause and suggested fix

In `cmd/pad/cmd_db.go`, the SQLite restore path calls `os.Open(inputFile)` and then `os.Create(dstPath)` without checking file identity. `os.Create` truncates the shared inode, after which `io.Copy` reads EOF and returns success.

Before `os.Create`, stat the source and existing destination and reject `os.SameFile(srcInfo, dstInfo)`. Add focused regression cases for direct same-path, symlink-alias, and hard-link-alias inputs, plus a distinct-file success control.

mattfaltyn:
## What does this PR do?

Refuses a SQLite restore before opening the destination when the backup and live database identify the same underlying file.

The guard uses `os.Stat` and `os.SameFile`, so it covers:

- the exact same path
- a symlink alias
- a hard-link alias

The existing distinct-file restore behavior is unchanged. Regression tests prove that every same-file form returns an error without altering the database, and that a distinct backup still copies successfully.

Closes #1331

## How to test

```bash
go test ./cmd/pad -run '^TestDBRestore(RejectsSameFile|CopiesDistinctFile)$' -count=1

# macOS: use the canonical temp path and skip two existing Linux /proc-only assertions
TMPDIR=/private/tmp GOFLAGS='-skip=^TestOwnerLiveness_(Pid|Socket)$' make test

GOOS=linux make lint
make test-nix-hash

# With Node 24, matching .nvmrc and CI
make build
cd web && npm run check && npm run test
```

Local results:

- focused restore tests: pass
- full locally applicable Go suite: pass
- Linux-target golangci-lint suite: 0 issues
- Nix hash script tests: 60 passed, 0 failed
- full build: pass
- Svelte check: 0 errors
- Vitest: 172 files and 2,464 tests passed

Host note: an unmodified `make test` on macOS runs two assertions in `internal/cli/session_owner_test.go` that expect Linux `/proc` process-start tokens; both return `unknown` on non-Linux by design. An unmodified macOS `make lint` similarly sees the Linux-only `errProcStatMalformed` sentinel as unused. The commands above exercise every locally applicable test and run the full lint suite for the same Linux target as CI. No unrelated platform-test changes are included here.

## Checklist

- [x] `make build` passes
- [x] `make test` passes for all locally applicable tests (host exception documented above)
- [x] Regression tests added
- [x] TypeScript types updated (not applicable; no API change)
- [x] CLI help text updated (not applicable; no command or flag change)

### DIFF
--- a/cmd/pad/cmd_db.go
+++ b/cmd/pad/cmd_db.go
@@ -220,8 +220,12 @@ WARNING: This will overwrite the current database contents.`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			inputFile := args[0]
-			if _, err := os.Stat(inputFile); os.IsNotExist(err) {
-				return fmt.Errorf("backup file not found: %s", inputFile)
+			inputInfo, err := os.Stat(inputFile)
+			if err != nil {
+				if os.IsNotExist(err) {
+					return fmt.Errorf("backup file not found: %s", inputFile)
+				}
+				return fmt.Errorf("stat backup file: %w", err)
 			}
 
 			dbDriver := os.Getenv("PAD_DB_DRIVER")
@@ -274,6 +278,13 @@ WARNING: This will overwrite the current database contents.`,
 				return fmt.Errorf("load config: %w", err)
 			}
 			dstPath := cfg.DBPath
+			if dstInfo, err := os.Stat(dstPath); err == nil {
+				if os.SameFile(inputInfo, dstInfo) {
+					return fmt.Errorf("backup file %s is the SQLite database being restored; choose a different backup file", inputFile)
+				}
+			} else if !os.IsNotExist(err) {
+				return fmt.Errorf("stat database: %w", err)
+			}
 
 			// Refuse to overwrite the database out from under a running server:
 			// the server holds it open and its background WAL checkpointer could
--- a/cmd/pad/cmd_db_restore_test.go
+++ b/cmd/pad/cmd_db_restore_test.go
@@ -1,0 +1,81 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestDBRestoreRejectsSameFile(t *testing.T) {
+	const contents = "database contents"
+
+	aliases := []struct {
+		name string
+		path func(t *testing.T, live string) string
+	}{
+		{"same path", func(_ *testing.T, live string) string { return live }},
+		{"symlink", func(t *testing.T, live string) string {
+			link := filepath.Join(filepath.Dir(live), "backup.db")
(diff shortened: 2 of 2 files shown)


xarmian:
Merged, thanks. The report on #1331 was the kind that makes the fix easy to take: three reproductions with hashes and a control run, and the fix itself uses os.SameFile instead of comparing path strings, which is the right check for this. This is going out as v0.15.2 today.

